PR #6, opened 2018-03-13
Proposed modifications for the documentation

dogoepp:
In addition to these commits, a few comments and questions on the guide.md:

- Robot Class / Controllers
  - what is an active controller ?
  - how to know which ID was given to a newly registered controller ?
  - what does reinitialisation of a controller mean ?
  - maybe we need to explain how weighting behaves and quickly say that it might be nice to have more than one controller (if it's true)
- Robot Class / Other functionalities
  - I don't get what "attached damages" is
- RobotControl Class / Main methods / Constructor
  - we now have method prototypes whereas we used to have usage examples; I prefer the former
- RobotDARTSimu Class / Robots: since `num_robots` retunrs a `size_t`, I think that we should enforce a limit on the number of robots
- RobotDARTSimu Class / Descriptors: what is `desc_dump` ?

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -298,7 +298,7 @@ std::shared_ptr<robot_dart::Robot> robot(size_t index) const;
 
 **Descriptors**
 
-In *RobotDARTSimu* there's the possibility of attaching **descriptors**. *Descriptors* are functors that can be used either to log information/data (e.g., state trajectories) or to trigger a stop in the simulation (e.g., a NaN value or safety reasons). All descriptors must **inherit** from `robot_dart::descriptor::BaseDescriptor`:
+In *RobotDARTSimu* there's the possibility of attaching **descriptors**. *Descriptors* are functors that can be used to log information/data (e.g., state trajectories) or to stop the simulation (e.g., a NaN value was produced or for safety reasons). All descriptors must **inherit** from `robot_dart::descriptor::BaseDescriptor`:
 
 ```cpp
 struct BaseDescriptor {
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -36,7 +36,7 @@ auto my_robot = std::make_shared<robot_dart::Robot>(
 
 ### Controllers
 
-In each robot you can have attached multiple controllers with different weights. Here's the basic functionality:
+Multiple controllers can be attached to a robot, each with their own weight. Here's the basic functionality:
 
 ```cpp
 std::shared_ptr<robot_dart::Robot> my_robot = ...;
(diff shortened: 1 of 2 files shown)


costashatz:
Hey,

Thanks for the effort! Your 

> what is an active controller ?

You are right about this one. An active controller is one that was properly initialized (i.e., well-behaved controller parameters) and that will be used when updating the robot (i.e., at each time-step). I will make it clearer. I will also make it easy to make a controller inactive on demand.

> how to know which ID was given to a newly registered controller ?

There's no ID; that's why I am saying index. Basically, each controller is added in the end of a vector and you can access this vector by indices. I agree with you that is not the best way to do it, but using an `std::unordered_map` seems overkill for what I want (it'd be very rare having more than 2-3 controllers and usually they will not change during the lifetime of a robot). I will try to think of a better solution..

> what does reinitialisation of a controller mean ?

It tries to reconfigure the controller. This is used along with the `active` thing. This is used in 2 cases:

- When the robot changes structure (i.e., when `robot->fix_to_world()` or `robot->free()` are called)
- When a controller had some wrong parameters (thus inactive), we updated the parameters and thus we want to re-initialize it to make it active.

I'll put something in the docs..

> maybe we need to explain how weighting behaves and quickly say that it might be nice to have more than one controller (if it's true)

In general I agree with you but it's not the most important feature of the library and I have too many things in my TO-DO list (that are more important)..

> I don't get what "attached damages" is

Me neither :stuck_out_tongue: .. I will add a small section about damages (or remove them completely)..

> we now have method prototypes whereas we used to have usage examples; I prefer the former

You're right! I'll change it..

> since num_robots retunrs a size_t, I think that we should enforce a limit on the number of robots

I cannot really imagine a simulation in DART with that many robots that a `size_t` cannot fit! I think we should be good! :smile: 

> what is desc_dump ?

From the guide.md:

*desc_dump defines the number of steps at which the descriptor should be called*

dogoepp:
Thanks for your replies.

> There's no ID; that's why I am saying index. Basically, each controller is added in the end of a vector and you can access this vector by indices. I agree with you that is not the best way to do it, but using an std::unordered_map seems overkill for what I want (it'd be very rare having more than 2-3 controllers and usually they will not change during the lifetime of a robot). I will try to think of a better solution..

My bad, index. However, could it be useful if `add_controller` returned the index of a newly added controller ? That's not perfect, but a middle ground.

> From the guide.md:
>> desc_dump defines the number of steps at which the descriptor should be called

If I pretend not to have heard your oral explanation, I don't get what this means. What you intend to say is something like "the descriptor will be called every `dexc_dump` iteration of the simulation."

costashatz:
> we now have method prototypes whereas we used to have usage examples; I prefer the former

You mean you prefer usage examples right?

dogoepp:
I find that the usage examples tend to lack information on return type of a function. If all your functions return void, it doesn't matter.

costashatz:
> I find that the usage examples tend to lack information on return type of a function. If all your functions return void, it doesn't matter.

So you prefer the method prototypes.. Agreed!

PR #9, opened 2018-06-01
Implemented a simpler version of the PD controller

lothas:
Implemented a simpler version of the PD controller


costashatz:
@lothas Thanks!

PR #10, opened 2018-06-01
Improved the simulation time check to make sure it doesn't keep going…

lothas:
… due to an insignificant time difference.


costashatz:
Thanks again!

PR #19, opened 2019-05-21
fix a leak from DART, this code should be compatible even after DART'…

### LINKED ISSUE #18: Memory leak with the new off-screen cameras
When using the cameras in sferes (QD algorithms), the memory consumption rises linearly with the number of evaluation (while it should remain more or less constant at some point). 
To give an order of magnitude I reach 20GB of ram usage in less than 100 generations (pop size =200). 

I am investigating the issue, but if someone has some ideas or hints, they are welcome.

Aneoshun:
fixes #18

### DIFF
--- a/src/robot_dart/graphics/base_graphics.hpp
+++ b/src/robot_dart/graphics/base_graphics.hpp
@@ -11,6 +11,8 @@ namespace robot_dart {
 
             BaseGraphics(const dart::simulation::WorldPtr&) {}
 
+            virtual ~BaseGraphics() {}
+
             virtual bool done() const
             {
                 return false;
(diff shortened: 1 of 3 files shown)


Aneoshun:
Just for reference, here is the output of valgrind massif after running sferes an a camera doing nothing (not even taking a shot) for 500 evaluations:
Screenshot 2019-05-21 at 17 50 21
The same code with the fix:
Screenshot 2019-05-21 at 18 09 34

The fix makes the memory consumption constant (as expected)!

costashatz:
Looks great! Thanks @Aneoshun!! **Merging..**

PR #24, opened 2019-07-31
add function add_grid in graphics

Aneoshun:
Hi @costashatz,

This is a simple pull request to add the possibility to add a grid on the ground in the visualisation. 
I have added this grid in the `hexapod` example to show the result. 

Let me know what you think. 

Best,

### DIFF
--- a/src/robot_dart/graphics/graphics.hpp
+++ b/src/robot_dart/graphics/graphics.hpp
@@ -80,7 +80,15 @@ namespace robot_dart {
                     osg::Vec3d(_camera_pos(0), _camera_pos(1), _camera_pos(2)), osg::Vec3d(_look_at(0), _look_at(1), _look_at(2)), osg::Vec3d(_camera_up(0), _camera_up(1), _camera_up(2)));
                 _osg_viewer->home();
             }
-
+	  
+	    void add_grid(double size = 1, size_t number = 10)
+	    {
+	        osg::ref_ptr<dart::gui::osg::GridVisual> grid = new dart::gui::osg::GridVisual();
+		grid->setNumCells(number);
+		grid->setMinorLineStepSize(size);
+		_osg_viewer->addAttachment(grid);
+	    }
+	  
             void enable_default_lights(bool enable = true)
             {
                 _osg_viewer->switchHeadlights(enable);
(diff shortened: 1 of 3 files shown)


src/robot_dart/graphics/graphics.hpp
```diff
@@ -80,7 +80,15 @@ namespace robot_dart {
                     osg::Vec3d(_camera_pos(0), _camera_pos(1), _camera_pos(2)), osg::Vec3d(_look_at(0), _look_at(1), _look_at(2)), osg::Vec3d(_camera_up(0), _camera_up(1), _camera_up(2)));
                 _osg_viewer->home();
             }
-
+	  
+	    void add_grid(double size = 1, size_t number = 10)
+	    {
+	        osg::ref_ptr<dart::gui::osg::GridVisual> grid = new dart::gui::osg::GridVisual();
```
costashatz: Thank you for this PR! Are we sure that every DART version >= 6 (that we support with robot_dart) has `osg::GridVisual`? I have a feeling that this was introduced quite lately (I might be wrong)..

Aneoshun:
Hum... you're right. To maintain compatibility, but still enabling me to add grid in my experiments, added a getter to access the _osg_viewer (this way I can add the grid in my exp files). 

Does this sound good for you?

costashatz:
> Does this sound good for you?

Could we do it with an ifdef depending on the version? Similarly to the clone function (see [here](https://github.com/resibots/robot_dart/blob/master/src/robot_dart/robot.cpp#L51-L55))?

If not, this is okay also..

Aneoshun:
I can actually do both (the getter might always be useful)

costashatz:
> I can actually do both (the getter might always be useful)

Good point!

Although I am working on the Magnum integration and seems to be working quite nicely with several threads etc... And it's much more modular than OSG.. So I'd most probably switch to this one; although the open-scene graph integration will remain since DART supports it by default..

Aneoshun:
You can rename this "graphics" into "osg_graphics" and have the same for Magnum. Personally, I don't have much preference or need, I am just playing with the tools at hand. So if you move robot_dart to Magnum, I will be happy to migrate there as well at some point.

costashatz:
> You can rename this "graphics" into "osg_graphics" and have the same for Magnum. Personally, I don't have much preference or need, I am just playing with the tools at hand. So if you move robot_dart to Magnum, I will be happy to migrate there as well at some point.

I have created a new namespace (named "gui") and kept the graphics with aliases to the OSG integration: see [here](https://github.com/resibots/robot_dart/tree/magnum/src/robot_dart/gui)..

I will let you know once I finish it (not too much time at the moment)..

Aneoshun:
I put the `add_grid` function back, with preprocessor to check the right version of dart

costashatz:
@Aneoshun thanks for this PR again!

Minor changes again (mostly formatting; clang complains :P) and adding this as an example to the newly added meshes example that is more visible than the hexapod (that already has a floor)..

If everything works on your side, I will merge..

Aneoshun:
This is working perfectly on my side. You can merge.

PR #25, opened 2019-07-31
change visuals of materials

Aneoshun:
Hi again @costashatz, 

Another very small pull request. We are using a new urdf with meshes and we noticed that the color (defined in the dae file) was not displayed in our visualisations. Changing `SHAPE_COLOR` to `MATERIAL_COLOR` solved our issues, without changing the colors of all the examples in robot_dart. 

Can this small changed be merged to master?

Best,

### DIFF
--- a/src/examples/meshes.cpp
+++ b/src/examples/meshes.cpp
@@ -20,6 +20,10 @@ int main()
 
     global_robot->fix_to_world();
     global_robot->set_position_enforced(true);
+    // Use the material information coming from the meshes instead of the URDF file
+    // global_robot->set_color_mode(dart::dynamics::MeshShape::ColorMode::MATERIAL_COLOR);
+    // Use the material information from the meshes only for a specific BodyNode
+    // global_robot->set_color_mode(dart::dynamics::MeshShape::ColorMode::MATERIAL_COLOR, "iiwa_link_6");
 
     std::vector<double> ctrl;
     ctrl = {0., M_PI / 3., 0., -M_PI / 4., 0., 0., 0.};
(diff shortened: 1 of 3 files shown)


costashatz:
Thanks for the PR!

> Can this small changed be merged to master?

The default behavior should be choosing the color of the URDF (if specified). The latest versions of DART do this by default. So my proposition would be to create helpers to easily change all or individual `ShapeNodes`'s type of material visualizations (there are only 3 types: see [here](https://github.com/dartsim/dart/blob/master/dart/dynamics/MeshShape.hpp#L49)). In that way, when loading the URDF, we should keep the default behavior of DART and make easy one-liners to change if we want something different..

What do you think? Can you do this PR instead?

Aneoshun:
If I understood correctly what you meant, I have done it and it works well in my experiments.

costashatz:
@Aneoshun thanks again for this PR!

I changed a bit your code to do the following:

-  Keep the old behavior (by default `SHAPE_COLOR`) so that people already using the code, would not have unexpected behaviors.
- There's no point in having a member function that takes a `ShapeNode` and set the color mode since the user only has to call only function.
- I made a helper function for a `BodyNode` to make one liner a code that can potentially be 5-6 lines

If it works for you, let me know and I will merge..

Aneoshun:
There is a small issue with the current code: cloning (or copying) override the color changes. I think we should call set_color_mode only at the end of load_model (like it was before) not in the constructors.

Aneoshun:
Actually that's more complicated as set_color_mode uses _skeleton while load_model uses _tmp_skel. 

How do you want to proceed?

costashatz:
> How do you want to proceed?


fd8b1ea should do it! Let me know if it works for you..!

Aneoshun:
That's all good for me.

PR #30, opened 2019-10-28
Completely remove OSG GUI

costashatz:
This PR completely removes the OSG gui. Leaves Magnum as the only GUI available. ~~This is still WIP..~~

**EDIT:** This also replaces Sdl2Application with GlfwApplication which seems to be more stable.


costashatz:
@Aneoshun can you check if we need [this](https://github.com/resibots/robot_dart/blob/master/docs/cameras.md) with Magnum `WindowlessGLApplication` (just running the `magnum_contexts` example should suffice)? In my case, I didn't need it, but I want to be sure before removing it..

waf_tools/corrade.py
```diff
@@ -152,12 +152,10 @@ def find_in_string(data, text):
 
                 corrade_component_libpaths[component] = corrade_component_libpaths[component] + [lib_dir]
                 corrade_component_libs[component].append(lib)
-                if component == 'PluginManager':
-                    # PluginManager needs the libdl.so library
+                if component == 'PluginManager' or component == 'Utility':
+                    # PluginManager/Utility need the libdl.so library
```
mosra: minor: because PluginManager depends on Utility, only Utility need the `dl` lib, PluginManager will get it transitively :)

costashatz:
@Aneoshun @jbmouret let me know if this still works in Mac. Make sure that you update your installations (see the new installation.md in the `remove_osg` branch) and use the latest magnum master..

If this is still working on Mac, then I think we should be able to merge..

Aneoshun:
It does compile, but does not work. Apparently the shader compilation fails: 
```
[robot_dart WARNING]: "Shadows are not working properly on Mac! Disable them if you experience unexpected behavior.."
Platform::GlfwApplication: warning: the executable is not a HiDPI-enabled app bundle
GL::Shader::compile(): compilation of fragment shader 2 failed with the following message:
ERROR: 3:187: storage modifier qualifier 'const' must precede precision qualifiers
ERROR: 3:192: storage modifier qualifier 'const' must precede precision qualifiers
ERROR: 3:197: storage modifier qualifier 'const' must precede precision qualifiers
ERROR: 3:204: Use of undeclared identifier 'finalAmbientColor'
ERROR: 3:262: Use of undeclared identifier 'finalDiffuseColor'
ERROR: 3:270: Use of undeclared identifier 'finalSpecularColor'
ERROR: 3:273: Use of undeclared identifier 'diffuseReflection'
Assertion Magnum::GL::Shader::compile({vert, frag}) failed in ../src/robot_dart/gui/magnum/gs/phong_multi_light.cpp on line 35
Abort trap: 6
```
I pulled, and rebuilt Corrade, Magnum, Magnum-plugin, Magnum-integration and robot_dart.

Aneoshun:
fyi: 
```
OpenGL vendor string: ATI Technologies Inc.
OpenGL renderer string: AMD Radeon Pro 560X OpenGL Engine
OpenGL version string: 2.1 ATI-3.0.68
OpenGL shading language version string: 1.20
```

costashatz:
> OpenGL vendor string: ATI Technologies Inc.
> OpenGL renderer string: AMD Radeon Pro 560X OpenGL Engine
> OpenGL version string: 2.1 ATI-3.0.68
> OpenGL shading language version string: 1.20

This is a very old version of OpenGL (2.1). It's not possible to support it. We need OpenGL version bigger than 3.1...

Aneoshun:
According to: https://support.apple.com/en-us/HT202823
my MBP shoudl support OpenGL 4.1 and OpenCL 1.2...

costashatz:
> According to: https://support.apple.com/en-us/HT202823
> my MBP shoudl support OpenGL 4.1 and OpenCL 1.2...

Then why does it say OpenGL 2.1? I will check on a mac computer but this seems to me as a driver installation bug? Not related to robot_dart..

mosra:
Hahah oh my ... this is endless :see_no_evil: You'll get 2.1 on Mac if you don't request core GL 3.2+ at least -- their driver either support the old pre-GL3 compat profile or core GL3+, but not both. So this seems to be one such case.

There was some change in Catalina that might have broken this (I heard some reports that GL3.3 at least needs to be requested, otherwise you get 2.1, previously 3.2 was enough) -- did this work with Sdl2Application or is Sdl2App affected the same way? Random patch that *might* make things different:

```diff
diff --git a/src/Magnum/Platform/GlfwApplication.cpp b/src/Magnum/Platform/GlfwApplication.cpp
index ee9f3a17d..f76711c6e 100644
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -384,7 +384,7 @@ bool GlfwApplication::tryCreate(const Configuration& configuration, const GLConf
            compatibility contexts (which are the default). Unlike SDL2, GLFW
            requires at least version 3.2 to be able to request a core profile. */
         glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
-        glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2);
+        glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);
         glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
         glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, glFlags >= GLConfiguration::Flag::ForwardCompatible);
         #else

```

Aneoshun:
- Well, actually, I am not sure that I tested the shadows when we were still on SDL2, so I can't tell. 
- I tried your patch and I don't see any difference. 
- The version info that I pasted earlier come from glxinfo, which might be the most accurate source of info on OSX (e.g., it starts a XQuartz server...).

mosra:
Oh! What's the startup log from the engine, then? Does it show GL 2.1 as well? I don't trust `glxinfo`, especially on platforms where X11 is not native. GLFW uses CGL there, not GLX.

Aneoshun:
This is what the executable gives when I comment `Corrade::Utility::Debug magnum_silence_output{nullptr};  ` in graphics.hpp
```
git/robot_dart$ ./build/meshes
[robot_dart WARNING]: "Shadows are not working properly on Mac! Disable them if you experience unexpected behavior.."
Platform::GlfwApplication: warning: the executable is not a HiDPI-enabled app bundle
Renderer: AMD Radeon Pro 560X OpenGL Engine by ATI Technologies Inc.
OpenGL version: 4.1 ATI-3.0.68
Using optional features:
    GL_ARB_ES2_compatibility
    GL_ARB_separate_shader_objects
    GL_ARB_texture_filter_anisotropic
    GL_ARB_texture_storage
    GL_ARB_vertex_array_object
    GL_EXT_debug_label
    GL_EXT_debug_marker
Using driver workarounds:
    no-layout-qualifiers-on-old-glsl
GL::Shader::compile(): compilation of fragment shader 2 failed with the following message:
ERROR: 3:187: storage modifier qualifier 'const' must precede precision qualifiers
ERROR: 3:192: storage modifier qualifier 'const' must precede precision qualifiers
ERROR: 3:197: storage modifier qualifier 'const' must precede precision qualifiers
ERROR: 3:204: Use of undeclared identifier 'finalAmbientColor'
ERROR: 3:262: Use of undeclared identifier 'finalDiffuseColor'
ERROR: 3:270: Use of undeclared identifier 'finalSpecularColor'
ERROR: 3:273: Use of undeclared identifier 'diffuseReflection'
Assertion Magnum::GL::Shader::compile({vert, frag}) failed in ../src/robot_dart/gui/magnum/gs/phong_multi_light.cpp on line 35
Abort trap: 6
```

(I haven't reverted the patch introduced above)

costashatz:
In the shader, I am requiring OpenGL 3.2 and some additional features. Maybe that's the issue @mosra ? Should I require 3.3?

mosra:
Great! Then GL 2.1 was just a false alarm (and `glxinfo` is thus wrong). I expect it to work the same with or without the patch, considering I didn't get any recent reports about apps/examples refusing to work on Mac.

The remaining error is that `const` can be used for non-compile-time constants only since GLSL 4.20, and here you can get 4.1 at most. So removing the `const`s from here should make things work: https://github.com/resibots/robot_dart/blob/d8a81f77dbf275dbdd72d76d8180da1a2e11a4e1/src/robot_dart/gui/magnum/resources/PhongMultiLight.frag#L175-L189

@costashatz what's the reason for fully disabling magnum's startup log? it provides very useful info for debugging ;)

costashatz:
> @costashatz what's the reason for fully disabling magnum's startup log? it provides very useful info for debugging ;)

When running things in parallel with multiple contexts, I get the log information too many times! Of course, I could just disable it for each executable that needs to run in the cluster.. I will think about adding it back.. :)

costashatz:
@Aneoshun did 388b2d5 fix your issues?

Aneoshun:
@costashatz Yes it does! I now have a robot, with shadows and ground!

costashatz:
> @costashatz Yes it does! I now have a robot, with shadows and ground!

Great! Then I will merge soon! @jbmouret let me know if you have any comment/issue..

jbmouret:
The waf configure does not work for me anymore.
```
./waf configure    --controller=/Users/jmouret/git/resibots/hexapod_common/hexapod_controller --magnum_install_dir=/usr/local/include
Setting top to                           : /Users/jmouret/git/resibots/robot_dart 
Setting out to                           : /Users/jmouret/git/resibots/robot_dart/build 
Checking for 'clang++' (C++ compiler)    : /usr/bin/clang++ 
Checking for 'clang' (C compiler)        : /usr/bin/clang 
Checking boost includes                  : 1_71 
Checking boost libs                      : ok 
Checking for Eigen                       : /usr/local/include/eigen3 
Checking for DART includes (including io/urdf) : 6.10.0 in /usr/local/include 
Checking for DART libs (including io/urdf)     : ['dart', 'dart-utils', 'dart-utils-urdf'] 
DART: Checking for Assimp                      : /usr/local/include 
DART: Checking for Bullet Collision libs       : libs: ['LinearMath', 'BulletCollision', 'dart-collision-bullet'], bullet: /usr/local/include/bullet 
DART: Checking for Ode Collision libs          : Not found 
Checking for hexapod_controller includes       : /Users/jmouret/git/resibots/hexapod_common/hexapod_controller/include 
Checking for Corrade includes                  : /usr/local/include 
Getting Corrade configuration                  : ['BUILD_DEPRECATED', 'TARGET_UNIX', 'TARGET_APPLE'] 
Checking for Corrade components                : ['CorradePluginManager', 'CorradeUtility', '/usr/local/bin/corrade-rc'] 
Checking for Magnum includes                   : /usr/local/include 
Checking for Magnum lib                        : ['Magnum'] 
Getting Magnum configuration                   : ['BUILD_DEPRECATED', 'BUILD_MULTITHREADED', 'TARGET_GL'] 
Magnum: Checking for OpenGL includes           : /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/OpenGL.framework/Versions/A/Headers 
Magnum: Checking for OpenGL lib                : ['GL'] 
Magnum: Checking for MagnumGL lib              : ['MagnumGL'] 
Checking for Magnum components                 : Not found 
Magnum needs to be configured! Cannot proceed! 
Magnum needs to be configured! Cannot proceed! 
AVX compilation of dart                        : yes 
AVX compilation of dart-utils                  : yes 
AVX compilation of dart-utils-urdf             : yes 
-march=native (AVX support)                    : yes 
['-Wall', '-O3', '-g', '-faligned-new', '-march=native', '-std=c++14']
'configure' finished successfully (0.932s)
```

=> There is an issue with the magnum components (but magnum, corrade, etc. are OK). Any idea what is wrong? (the old 'magnum' branch binaries still work!). Also, would it be possible to print which component is missing? I guess only one of them is not detected.

costashatz:
> => There is an issue with the magnum components (but magnum, corrade, etc. are OK). Any idea what is wrong? (the old 'magnum' branch binaries still work!). Also, would it be possible to print which component is missing? I guess only one of them is not detected.

We no longer are using Sdl2, but Glfw3. Can you update your installation of Magnum following these instructions: https://github.com/resibots/robot_dart/blob/remove_osg/docs/installation.md#installing-magnum?

jbmouret:
Thanks! It works now.

src/robot_dart/gui/magnum/glfw_application.cpp
```diff
@@ -18,6 +19,7 @@ namespace robot_dart {
                 conf.setSize({static_cast<int>(width), static_cast<int>(height)});
                 conf.setWindowFlags(Configuration::WindowFlag::Resizable);
                 glConf.setSampleCount(8);
+                glConf.setVersion(Magnum::GL::Version::GL320);
```
mosra: I'm pretty sure this will cause NVidia drivers to be stuck at GL 3.2 (instead of going all the way to 4.6). Why was it needed? :)

costashatz: This is indeed true. But all we need is GL 3.2 + one additional feature. Will it affect performance if we do not use GL 4.6 (when available)?

mosra: No, but at some point you might want to opt-in into fancier features available only on newer versions (extensions depending on GL 4.0+ and such, for example). Right now there won't be any difference I think.

costashatz:
> Thanks! It works now.

**Merging then...** Thanks for all the feedback!

PR #33, opened 2019-11-21
Robust magnum detection

Aneoshun:
Hi, 

I have a small, yet annoying, issue when trying to compile robot_dart with python3 (i.e., without python2 on my system). The issue is when we look for the config files of magnum and corrade, which have an encoding that seems to be causing problem, in python3, not python2.

Here is a small PR that fix it. I have tested it on OSX, Ubuntu, with Python2 and Python3.

### DIFF
--- a/waf_tools/corrade.py
+++ b/waf_tools/corrade.py
@@ -7,6 +7,7 @@ Quick n dirty Corrade detection
 """
 
 import os
+import io
 from waflib import Utils, Logs
 from waflib.Configure import conf
 import copy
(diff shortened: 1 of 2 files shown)


costashatz:
> I have a small, yet annoying, issue when trying to compile robot_dart with python3 (i.e., without python2 on my system). The issue is when we look for the config files of magnum and corrade, which have an encoding that seems to be causing problem, in python3, not python2.
> 
> Here is a small PR that fix it. I have tested it on OSX, Ubuntu, with Python2 and Python3.

This is weird. In my arch linux, I have python3 enabled (python2 installed but not activated in any way; I checked and waf is calling python3) and everything works perfect. Any idea why is this happening? I guess it is Mac related?

Nevertheless, with your fixes everything is still the same and thus harmless. So I will merge tomorrow..

Aneoshun:
The issue with Python3 was on my Ubuntu singularity container, but I have no clue why it is behaving like this. With this fix, everything seems to work on my side.

costashatz:
> Ubuntu singularity container

This singularity container is a weird thing! :stuck_out_tongue: 

**Merging..**

Issue #42, opened 2020-05-10
Create a name to index map for Robots


costashatz:
We should create a mapping between names and indices of robots' joints/dofs for being able to query their states with the names instead of just indices.

PedroDesRobots:
We can implement this function like we did : 
```
std::map<std::string, int> bridge::dart_robot_loader::getIndexMap() {
  int numjoints = dRobot_->getNumJoints();
  std::vector<dart::dynamics::Joint*> joints;
  for (int i = 0; i < numjoints; ++i) {
    joints.push_back(dRobot_->getJoint(i));
    index_map_[joints[i]->getName()] = i;
  }
  return index_map_;
}
```

Then something like that :
```
int getIndex(std::string jointName){
  return index_map_(jointName);
}
```

costashatz:
@PedroDesRobots thanks for the code.. One thing to be careful is the fact that 1 joint does not mean one DoF. Many joints have zero-DoFs and one joint can have multiple-DoFs. We should discuss how to handle this. How do you see this? Another important thing is the mimic joints. I'll try to come up with a plan. If you have any ideas, let me know..

PedroDesRobots:
Yes sorry for the confusion.  In fact, `getForceTorque(joint_index)` uses
`auto jt = _skeleton->getJoint(joint_index)`. So it's joint related and not DoF.

In terms of physical quantities, we are always looking at the joint level. DoFs are more useful for geometrical aspects/ kinematics.

>  Another important thing is the mimic joints

What do you mean by mimic joints?

costashatz:
> What do you mean by mimic joints?

http://wiki.ros.org/urdf/XML/joint#Elements (search for <mimic>).

Joints that move according to another joint. Sometimes you have robots that have hardware mimicing (e.g. NAO humanoid), while in other cases it is quite useful for defining gripper joints (e.g. Talos humanoid).

These need extra caring while controlling them, because they appear as DoFs while you can only control one of them in practice (e.g., open/close gripper).

dalinel:
So maybe we could have one joint map for the controllable joints/dofs with some naming convention for the MultiDofs case and mimic joint case ?

And maybe another joint map with the non controllable and  zero-DoFs ?

costashatz:
> So maybe we could have one joint map for the controllable joints/dofs with some naming convention for the MultiDofs case and mimic joint case ?
> 
> And maybe another joint map with the non controllable and zero-DoFs ?

[...]

PR #43, opened 2020-05-10
Robot helper functions

### LINKED ISSUE #39: Add helper functions to Robot class
We should add helper functions to Robot class for people that are not very familiar with DART. Here's a list (should be updated accordingly):

- ~~get/set BasePose~~
- ~~get/set Positions~~
- ~~get/set Accelerations~~
~~- get/set Torques~~ We cannot set/get torques.
- ~~get/set Velocities~~
- ~~get ForceTorque~~
- ~~get/set LinearForces~~
- ~~get/set AngularForces~~
- ~~get LinkPositions~~
- ~~get/set LinkNames~~
- ~~get/set JointNames~~
- ~~get/set LinkMass~~
- ~~getCoM~~
- ~~addLinkMass~~

@jbmouret @dalinel @PedroDesRobots let me know if you have something to add here.. Most of these are straightforward to add..

costashatz:
This PR attempts to fix #39 ...

@jbmouret @dalinel @PedroDesRobots

### DIFF
--- a/src/examples/arm.cpp
+++ b/src/examples/arm.cpp
@@ -46,14 +46,14 @@ int main()
 #endif
     simu.add_descriptor(std::make_shared<StateDesc>(simu));
     simu.add_robot(g_robot);
-    std::cout << (g_robot->body_trans("arm_link_5") * size).transpose() << std::endl;
+    std::cout << (g_robot->body_pose("arm_link_5") * size).transpose() << std::endl;
     simu.run(2.5);
-    std::cout << (g_robot->body_trans("arm_link_5") * size).transpose() << std::endl;
+    std::cout << (g_robot->body_pose("arm_link_5") * size).transpose() << std::endl;
     ctrl = {0.0, -1.0, 1.5, -1.0};
     g_robot->controllers()[0]->set_parameters(ctrl);
     std::static_pointer_cast<robot_dart::control::PDControl>(g_robot->controllers()[0])->set_pd(20., 0.);
     simu.run(2.5);
-    std::cout << (g_robot->body_trans("arm_link_5") * size).transpose() << std::endl;
+    std::cout << (g_robot->body_pose("arm_link_5") * size).transpose() << std::endl;
 
     std::cout << std::static_pointer_cast<StateDesc>(simu.descriptor(0))->states.size() << std::endl;
 
--- a/src/examples/pendulum.cpp
+++ b/src/examples/pendulum.cpp
@@ -58,13 +58,13 @@ int main()
     // <Type>(desc_period)
     simu.add_descriptor<StateDesc>(2);
     simu.add_robot(g_robot);
-    std::cout << (g_robot->body_trans("pendulum_link_1") * size).transpose() << std::endl;
+    std::cout << (g_robot->body_pose("pendulum_link_1") * size).transpose() << std::endl;
     simu.run(2.5);
-    std::cout << (g_robot->body_trans("pendulum_link_1") * size).transpose() << std::endl;
+    std::cout << (g_robot->body_pose("pendulum_link_1") * size).transpose() << std::endl;
     ctrl = {2.5};
     g_robot->controllers()[0]->set_parameters(ctrl);
     simu.run(2.5);
-    std::cout << (g_robot->body_trans("pendulum_link_1") * size).transpose() << std::endl;
+    std::cout << (g_robot->body_pose("pendulum_link_1") * size).transpose() << std::endl;
 
     std::cout << std::static_pointer_cast<StateDesc>(simu.descriptor(0))->states.size() << std::endl;
 
--- a/src/python/robot.cpp
+++ b/src/python/robot.cpp
@@ -73,6 +73,10 @@ namespace robot_dart {
                 .def("base_pose", &Robot::base_pose)
                 .def("set_base_pose", &Robot::set_base_pose)
 
+                .def("num_dofs", &Robot::num_dofs)
+                .def("num_joints", &Robot::num_joints)
+                .def("num_bodies", &Robot::num_bodies)
+
                 .def("com", &Robot::com)
                 .def("com_velocity", &Robot::com_velocity)
                 .def("com_acceleration", &Robot::com_acceleration)
--- a/src/python/robot.cpp
+++ b/src/python/robot.cpp
@@ -90,7 +90,25 @@ namespace robot_dart {
                 .def("accelerations", &Robot::accelerations)
                 .def("set_accelerations", &Robot::set_accelerations)
 
+                .def("forces", &Robot::forces)
+                .def("set_forces", &Robot::set_forces)
+
                 .def("force_torque", &Robot::force_torque)
+
+                .def("set_external_force", (void (Robot::*)(const std::string&, const Eigen::Vector3d&, const Eigen::Vector3d&, bool, bool)) & Robot::set_external_force)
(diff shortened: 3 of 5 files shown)


costashatz:
@jbmouret @dalinel @PedroDesRobots 

I implemented most of the helper functions. We need to discuss the following:

- get ForceTorque (dof_index)
    - This ideally should return the force or torque that is related only to the specific DOF. In your dart_wrapper you are returning the full 6D wrench. I would return a double value with the force/torque only for a specific DOF (e.g., like what Frank or KUKA provides for each joint). How were you using this function?
- get/set LinearForces/AngularForces
    - No idea what you want to do with that. External forces?

costashatz:
We should also add the following helpers:

- getNumDofs
- getNumJoints
- getNumBodies

jbmouret:
For the force torque, I guess one of the use was to compute the 'energy' used by the motion.

For external forces, maybe to push the robot to see how it reacts?

costashatz:
> For the force torque, I guess one of the use was to compute the 'energy' used by the motion.

So we need the full force-torque? Good. I'll do this then..

> For external forces, maybe to push the robot to see how it reacts?

My question was whether what you meant is external forces or not..

PedroDesRobots:
`dartJointWrench getForceTorque(int DoF) `returns 6D Wrench of a body to get linear force and torque (Fx,Fy,Fz,Tx,Ty,Tz).

Then I'm using `Eigen::vector3d getLinearForce(std::string jointName) ` and `Eigen::vector3d getAngularForce(std::string jointName) ` to extract from the previous function only the Linear force or Torque (Angular Force).

I'm using also an another function `  std::map<std::string, int> getIndexMap() ` to link index and joint name, easier to use than putting the index number.

I'm currently using theses functions for the stabilizer based on feet F/T sensors feedback.

But It could also be used to mimic the torque sensor's feedback for Talos.

dalinel:
Hello,
Yes exactly this is what was recently added :

```
dartJointWrench bridge::dart_robot_loader::getForceTorque(int j) {
  dartJointWrench dWrench;

  Eigen::VectorXd F1(6), F2(6);
  F1.setZero();
  F2.setZero();

  dart::dynamics::Joint* joint = dRobot_->getJoint(j);
  dart::dynamics::BodyNode* dartParentBody = joint->getParentBodyNode();
  dart::dynamics::BodyNode* dartChildBody = joint->getChildBodyNode();
  // transform form body 1 to body 2
  Eigen::Isometry3d T12 = dartChildBody->getRelativeTransform();
  // transform from joint to body 2
  Eigen::Isometry3d TJ2 = Eigen::Isometry3d::Identity();
  TJ2.translation() = joint->getTransformFromChildBodyNode().translation();
  
  F2 = -dart::math::dAdT(TJ2, dartChildBody->getBodyForce());
  F1 = -dart::math::dAdInvR(T12, F2);

  dWrench.body1Force_ << F1(3), F1(4), F1(5);
  dWrench.body1Torque_ << F1(0), F1(1), F1(2);
  dWrench.body2Force_ << F2(3), F2(4), F2(5);
  dWrench.body2Torque_ << F2(0), F2(1), F2(2);

  return dWrench;
}

Eigen::Vector3d bridge::dart_robot_loader::getLinearForces(std::string joint_name){
  int index = index_map_[joint_name];
  dartJointWrench joint_force = getForceTorque(index);
  return joint_force.body1Force_;
}

Eigen::Vector3d bridge::dart_robot_loader::getAngularForces(std::string joint_name){
  int index = index_map_[joint_name];
  dartJointWrench joint_force = getForceTorque(index);
  return joint_force.body1Torque_;
}
```

costashatz:
> `dartJointWrench getForceTorque(int DoF) `returns 6D Wrench of a body to get linear force and torque (Fx,Fy,Fz,Tx,Ty,Tz).

Yes I just implemented that. See my latest commit. One thing to note is that you should not query the DoF but the joint. In DART a DoF does not have a parent link and a child link. Also one joint might have multiple DoFs (not very common in our robots, but can happen).

> I'm currently using theses functions for the stabilizer based on feet F/T sensors feedback.

This makes sense since this computation basically is giving what a force torque sensor would give.

> But It could also be used to mimic the torque sensor's feedback for Talos.

This is a bit more complicated, since we need to take into account the axis of each joint. We need to discuss about an API for the sensors in general.

> I'm using also an another function ` std::map<std::string, int> getIndexMap()` to link index and joint name, easier to use than putting the index number.

This is for another PR, look at #42...

> Yes exactly this is what was recently added

Thanks. I mostly did the same in my commit: see 91c7399.

costashatz:
@dalinel @PedroDesRobots @jbmouret can you have a look if we need to implement any other robot helper for the moment?

dalinel:
Maybe it would be too many helper functions
But we had functions making loops over what you implemented with
body_pose
body_name
joint_name

Something like this : 
```
std::vector<Eigen::Isometry3d> bridge::dartRobotSim::getLinkPositions() {
  std::vector<Eigen::Isometry3d> tf;
  auto bodyNodes = dRobot_->getBodyNodes();
  for (int i = 0; i < bodyNodes.size(); ++i) {
    tf.push_back(bodyNodes[i]->getWorldTransform());
  }
  return tf;
}

std::vector<std::string> bridge::dartRobotSim::getLinkNames() {
  std::vector<std::string> linkNames;
  auto bodyNodes = dRobot_->getBodyNodes();
  for (int i = 0; i < bodyNodes.size(); ++i) {
    linkNames.push_back(bodyNodes[i]->getName());
  }
  return linkN

std::vector<std::string> bridge::dartRobotSim::getJointNames() {
  std::vector<std::string> jointNames;
  int numjoints = dRobot_->getNumJoints();
  std::vector<dart::dynamics::Joint*> joints;
  for (int i = 0; i < numjoints; ++i) {
    joints.push_back(dRobot_->getJoint(i));
    jointNames.push_back(joints[i]->getName());
  }
  return jointNames;
}
```
And also for the body_mass a function that add a weight to the already set one 
```
void bridge::dartRobotSim::addLinkMass(std::string linkName,
                                            double addMass) {
  double prevMass = dRobot_->getBodyNode(linkName)->getMass();
  dRobot_->getBodyNode(linkName)->setMass(prevMass + addMass);
}
```

What do you think about adding those ?

costashatz:
> But we had functions making loops over what you implemented with

What was the usage of these functions? I do not see why they will be useful. Maybe the joint_names and link_names could be useful to get all the possible values. I'll do those two then..

> And also for the body_mass a function that add a weight to the already set one

Why not? This is harmless..

costashatz:
> Maybe it would be too many helper functions

@dalinel Done with 7e9b3da.

costashatz:
I think that we can **merge** this.. @jbmouret @dalinel @PedroDesRobots let me know if you have any objection/issue..

dalinel:
Yes it looks good, I am taking this to start the joint_map implementation

PedroDesRobots:
We could maybe add the function `joint_index(std::string joint_name)`, because when we want to use the function `force_torque(size_t joint_index)` is not trivial to know which index corresponds to which joint name.

@dalinel is what you want to add?

costashatz:
> We could maybe add the function `joint_index(std::string joint_name)`, because when we want to use the function `force_torque(size_t joint_index)` is not trivial to know which index corresponds to which joint name.
> 
> @dalinel is it what you want to add?

Good catch! I guess @dalinel can add this in the new PR..

Issue #48, opened 2020-05-11
Add helper functions for RobotDARTSimu


costashatz:
Add a few helper functions for RobotDARTSimu:

- Selecting the collision detector with a string and not DART code
- Add/Remove collision filter (mask collisions) without any DART code
- ~~Custom collision checks, and casting of rays~~ This is too much for now; I will make a different issue for this one.
- ...

@jbmouret @dalinel @PedroDesRobots let me know if you need more functions for the main simulator object. Recording videos/data of simulation will go in a different issue/PR..

costashatz:
Fixed by #56...

PR #56, opened 2020-05-14
RobotDARTSimu helpers

costashatz:
This PR attempts to partly address #48.. I did not do the custom collision checks and ray casting because we would need to define a structure for the results. Since we do not need it at the moment, I will handle this at the moment.

The collision filtering is done through a bitmask check.

### DIFF
--- a/src/python/simu.cpp
+++ b/src/python/simu.cpp
@@ -46,6 +46,7 @@ namespace robot_dart {
                 .def("num_robots", &RobotDARTSimu::num_robots)
                 .def("robots", &RobotDARTSimu::robots)
                 .def("robot", &RobotDARTSimu::robot)
+                .def("robot_index", &RobotDARTSimu::robot_index)
 
                 .def("add_robot", &RobotDARTSimu::add_robot, py::keep_alive<2, 1>())
                 .def("remove_robot", (void (RobotDARTSimu::*)(const std::shared_ptr<Robot>&)) & RobotDARTSimu::remove_robot)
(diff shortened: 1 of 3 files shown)


costashatz:
@jbmouret @dalinel @PedroDesRobots let me know if you like this PR and I will merge.. The collision filtering becomes something like:

```cpp
simu.set_collision_mask(simu.robot_index(arm), "arm_link_name", mask); // for bodies/links
simu.set_collision_mask(simu.robot_index(second_arm), "second_arm_link_name", mask);

simu.set_collision_mask(simu.robot_index(arm), mask); // for robots; this will apply the same mask to all BodyNodes of the robot
simu.set_collision_mask(simu.robot_index(second_arm), mask);
```

All the bodies with the same mask will not collide (comparison is bit-wise).

dalinel:
Sounds good !
Just a question, why did you choose to use this bitmask form instead of the default ignoresCollision ?

costashatz:
> why did you choose to use this bitmask form instead of the default ignoresCollision ?

I guess you mean the `addBodyNodePairToBlackList()` because `ignoresCollision()` is a generic function of the `CollisionFilter` class (which I am using). The reason for choosing the bitmask is that I find it more generic. I set each of my robots with a specific mask each and then the filtering determines on each own who gets collided or not. In that way, we can more easily handle dynamically changing collision filters; we just need to change on binary number. Also with the bitmasks we can do much more advanced checks than just colliding pairs (remember that the check is mask1&mask2==0, this means that if mask1=1001 and mask2 is 1000, we do not get zero; exploiting this one can create some quite complex collision-ignoring logics involving many "groups" of objects that collide or not with other "groups" of objects etc.). In addition, the bitwise operations are much faster than any other operations, etc..

dalinel:
Yes this is what I meant and thanks for the explanation !

costashatz:
**Merging..**

PR #57, opened 2020-05-15
Adding ghost robots and other visualizations

### LINKED ISSUE #40: Add API for easily adding markers/ghost robots and other visualizations
We should add helper functions to RobotDARTSimu for easily adding visualization things. Here's a list:

- Ghost robots (clone of robots with different colors/shading that do not interact with the environment and are there just for visualization purposes) --- maybe this should go to Robot class
- Add visualization spheres
- Add visualization cubes/rectangles
- ...

@jbmouret @dalinel @PedroDesRobots let me know if you have something to add..

costashatz:
This PR fixes #40.. ~~Still **WIP**..~~ This will break the GUI API slightly again..

### DIFF
--- a/src/examples/arm.cpp
+++ b/src/examples/arm.cpp
@@ -42,7 +42,7 @@ int main()
 
     robot_dart::RobotDARTSimu simu;
 #ifdef GRAPHIC
-    simu.set_graphics(std::make_shared<robot_dart::gui::magnum::Graphics<>>(simu.world()));
+    simu.set_graphics(std::make_shared<robot_dart::gui::magnum::Graphics<>>(&simu));
 #endif
     simu.add_descriptor(std::make_shared<StateDesc>(simu));
     simu.add_robot(g_robot);
--- a/src/examples/cameras.cpp
+++ b/src/examples/cameras.cpp
@@ -8,8 +8,8 @@
 
 class MyApp : public robot_dart::gui::magnum::GlfwApplication {
 public:
-    explicit MyApp(int argc, char** argv, const dart::simulation::WorldPtr& world, const robot_dart::gui::magnum::GraphicsConfiguration& configuration = robot_dart::gui::magnum::GraphicsConfiguration())
-        : GlfwApplication(argc, argv, world, configuration) {}
+    explicit MyApp(int argc, char** argv, robot_dart::RobotDARTSimu* simu, const robot_dart::gui::magnum::GraphicsConfiguration& configuration = robot_dart::gui::magnum::GraphicsConfiguration())
+        : GlfwApplication(argc, argv, simu, configuration) {}
 
 protected:
     void keyPressEvent(KeyEvent& event) override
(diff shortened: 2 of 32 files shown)


costashatz:
At the moment we can clone robots in `ghost` mode: the user can choose the color of the ghost robot and ghost robots have no shadows. Here's an example:
image

I am also thinking to make it possible to select whether a camera should render ghost robots and other visualizations. For example, we might want the main camera to display the ghost robots/visualizations for inspection, but the sensor cameras to not display them.

My thoughts are as follows:

*Do we really need this? I mean we would usually use the main camera only when debugging. When running the learning algorithms or other simulation, it will be rare that we would want a gui visualization as this will slow things down. So one can create 2 executables with defines: one for inspection and another for running without the GUI. On the other hand, being able to choose which objects get displayed and which not might be useful in other cases as well.*

@jbmouret @dalinel @PedroDesRobots let me know what you think.. I will add the rest of the visualizations as well (spheres, cubes, etc.)..

jbmouret:
On our side, we are not using cameras as sensors, so this does not make any difference for now. But the idea of 'ghosts' is that they do not appear on sensors, so we might want to keep this in mind anyway.

costashatz:
> But the idea of 'ghosts' is that they do not appear on sensors

It was much easier to do than I thought! I pushed the fix. No we can choose whether we want to draw ghosts objects, both in the main camera and every other sensor..

costashatz:
@jbmouret @dalinel @PedroDesRobots this should finished. I will not include the ConvexHull in this iteration... Let me know if I can merge..

dalinel:
It is ok for me !

costashatz:
@Aneoshun @Lookatator pinging you in case this is breaking your code (we are breaking slightly the GUI API)..

Aneoshun:
Hey, that's a fantastic feature! Thanks. 
If it works for you, you can merge. We have pined the commit SHA of all the modules used and we will remain a bit behind master, while our current problems with dart are not solved.

costashatz:
> Hey, that's a fantastic feature! Thanks.
> If it works for you, you can merge. We have pined the commit SHA of all the modules used and we will remain a bit behind master

Great! Thanks for the information!

> while our current problems with dart are not solved

What problems are you facing with DART? Or you meant tasks?

Aneoshun:
The "memory leak" fix introduced last week causes some issues in thread safety with random segfaults during sferes experiments.
https://github.com/dartsim/dart/issues/1468

costashatz:
> The "memory leak" fix introduced last week causes some issues in thread safety with random segfaults during sferes experiments.
> [dartsim/dart#1468](https://github.com/dartsim/dart/issues/1468)

@Aneoshun thank you for letting us know. Thankfully we do not really depend on the latest master for robot_dart, but it's good to know. Can you make a separate issue for this? Thanks again!

Aneoshun:
A separate issue, where? On Robot_dart, just to keep track of the issue in dart?

costashatz:
> On Robot_dart, just to keep track of the issue in dart?

yes thank you!

Aneoshun:
done

PR #58, opened 2020-05-15
Save a video

jbmouret:


jbmouret:
@costashatz do we keep the FPS fixed?

costashatz:
I tried your code and it is working on my Ubuntu machine. Let me try also in my arch-linux machine and let you know. It's nice that we can record videos now!!

> @costashatz do we keep the FPS fixed?

We need to make this better in general. So for this PR, let's keep it like this and we should open a different issue/PR for fixing the FPS things.

jbmouret:
I don't know about redirecting the output of ffmpeg: this might be useful to have some log when things go wrong (not mandatory, but silently failing is always frustrating).

Also we could check that ffmpeg is in the path during the waf configuration.

costashatz:
> I don't know about redirecting the output of ffmpeg: this might be useful to have some log when things go wrong (not mandatory, but silently failing is always frustrating).

ffmpeg produces too much output! Imagine that we have multiple cameras recording videos! It'll be a mess! We could make this optional somehow.

> Also we could check that ffmpeg is in the path during the waf configuration.

Do we really need this? Not having ffmpeg is not breaking anything and will just ignore the calling of `record_video` (that outputs a warning).

jbmouret:
OK for being not-verbose for now. We do not really need the waf check, but it might be a simpler call to the right function of waf (one line), which I do not remember (checking that a binary is in the path is certainly part of the standard functions).

costashatz:
@jbmouret I made the copying of the image data faster. Can you verify that this is still working for you?

costashatz:
@jbmouret on my arch-linux machine, I get (it's not specific to the pendulum example, I get this for all examples): 

```bash
/usr/bin/ld: src/examples/pendulum.cpp.6.o: undefined reference to symbol 'pthread_join@@GLIBC_2.2.5'
/usr/bin/ld: /usr/lib/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

I guess we are forgetting to link to a Boost library that includes pthread?

jbmouret:
Yes, I guess so. We should add boost.process
(what is missing is simply a -lpthread, however).

costashatz:
> We should add boost.process

There is no boost process library.

> (what is missing is simply a -lpthread, however).

Yes. I just added pthread as a lib. Maybe we should do this only if we want the video?

costashatz:
@jbmouret what is left on this? (apart from merging master to resolve conflicts) I guess relying on `pthread` is totally ok, right?

jbmouret:
I think we can merge.

costashatz:
> I think we can merge.

Found a small bug and I fixed it. I will wait for the CI to build/pass and then I merge.

costashatz:
@jbmouret apparently `Boost.process` is available from Boost >=1.64. Ubuntu Xenia (16.04)l has Boost 1.58. How should we handle this? Check the version and disable the video recording if we do not have at least 1.64?

jbmouret:
Yes, this sounds reasonable. 16.04 is old now, so no need to do more.

jbmouret:
(of course, we have a 16.04 on our own cluster...).

costashatz:
**Merging...**

PR #60, opened 2020-05-16
CI & Tests

### LINKED ISSUE #20: Add CI & unit tests
Add continuous integration and unit tests..

costashatz:
This PR will fix #20.. I have already implemented:

- CI with travis with several alternatives
- Basic testing skeleton using Boost unit tests

It remains:
- [x] CI in multiple Ubuntu distros, and OS X
- [x] Improve DART compilation in CI, see [here](https://github.com/dartsim/dart/blob/master/.ci/install_linux.sh)
- [x] CI for python bindings
- [x] Meaningful tests; this one I will create one PR per group of tests to be able to track them well


costashatz:
@jbmouret how can I install DART and Magnum in Mac with brew? I want to make one test scenario there too..

jbmouret:
I did not install dart & magnum with brew (I compiled everything). I installed the deps (fcl).

jbmouret:
Do we really want to use the master of the deps in the CI? It would be safer to use some release or specific commit (I agree that it is interesting to know that the dep broke our code, but we should not ask our user to install the master...).

costashatz:
> Do we really want to use the master of the deps in the CI?

I have everything. Stable releases and master.. It's good to know when something from master of dart/magnum breaks our code, because it'll ping us to update it. Otherwise we lose contact..

> we should not ask our user to install the master

We should not ask our users to install the master. I am taking the stable versions from the packages. In 2 different Ubuntu versions and now I am doing Mac. I found many small bugs and corrected them. I just also try the master because I always like to have the latest versions (I am using arch-linux as my everyday desktop).

costashatz:
@jbmouret I think this is done. We have the following variants for testing:

**Mac**:
- DART latest stable version from brew, 
- DART latest stable version from brew, Magnum latest master from brew

**Ubuntu Bionic**:
- Test python bindings with latest DART (master)
- Test CMake with latest stable version of DART (ppa) with and without Magnum
- Test with latest DART master with and without Magnum

**Ubuntu Xenial**:
- DART latest stable release (ppa) with and without Magnum

Some remarks:
- DART from v6.8 and on, only supports Ubuntu Xenial or later; I guess we should require DART >=v6.8
- Travis does not yet have the latest Ubuntu distros, but there things should work easier
- We still depend on Magnum master, but then next Magnum release should be available somewhere around end of June this year, and then we should stick to this version (I would be updating the code, but keeping backwards compatibility)
- I will open new PRs for the actual tests. Now I am only doing some dummy tests for the robot class..

costashatz:
**Merging** to be able have at least compilation verification in different setups already. I will create new PRs for unit tests.

PR #61, opened 2020-05-18
Fix compilation on mac with ODE collision

jbmouret:
On my mac, the libode was compiled in static (libode.a) but we were only checking for the .dynlib library.

It broke the compilation since Dart was actually compiled with ODE but we were not finding ODE.


costashatz:
**Thanks! Merging...**

PR #62, opened 2020-05-18
working on #58 (separate build for examples)

jbmouret:
It does not work yet (RobotDARTSimu is not linked)

### DIFF
--- a/wscript
+++ b/wscript
@@ -260,7 +260,7 @@ def build(bld):
 
     #### installation of the cmake config (waf install)
     prefix = bld.get_env()['PREFIX']
-    # config
+    # CMAKE config
     with open('cmake/RobotDARTConfig.cmake.in') as f:
         defines_magnum = ''.join((x + ';').replace('"', '\\"') for x in bld.get_env()['DEFINES_Magnum'])
         magnum_libs = ''.join(x + ';' for x in bld.env['magnum_libs'].split(' '))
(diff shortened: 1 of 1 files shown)


jbmouret:
It seems to me that the build context does not include RobotDARTSimu (since this is a different build object) and therefore we cannot use the libraries... we could simply do `./waf --examples` but I am not sure we can do `./waf examples'

In Limbo, it worked but we had no link, I think. But maybe there is another issue...

costashatz:
We need dependencies: see [here](https://waf.io/book/#_dependencies).

jbmouret:
What I did (last commit) adds the dependency, right?

costashatz:
> What I did (last commit) adds the dependency, right?

I guess so. It compiles on my arch-linux system. I will try also the Ubuntu..

costashatz:
@jbmouret this is working on all my machines. Should we **merge**?

jbmouret:
I wanted to first remove the redundancy between the graphics & non-graphics version (make a loop instead).

jbmouret:
Should we keep the hexapod example? I like hexapods but depending on the external controller makes things complex for nothing.

We could also keep it and adds the file for its controller (not depending on the resibots controller). It should be a single file that could be in the example.

costashatz:
> Should we keep the hexapod example? I like hexapods but depending on the external controller makes things complex for nothing.

Let's remove it.. It doesn't really add anything to `robot_dart`. We can keep a hexapod example where it does not move and just stays in place with `SERVO` actuators. What do you think?

jbmouret:
I am all for removing things!

jbmouret:
(can you check the last push? we can merge after this)

costashatz:
> (can you check the last push? we can merge after this)

Did you remove the hexapod example?

jbmouret:
Not yet. We could do a second PR.

costashatz:
> (can you check the last push? we can merge after this)

It's working.. We can **merge**..

PR #65, opened 2020-05-19
remove the dependency to hexapod_controllers

jbmouret:
This was only for an example and not widely used.


wscript
```diff
@@ -333,29 +329,9 @@ def build_examples(bld):
                             includes = './src',
                             uselib = 'PTHREAD ' + bld.env['magnum_libs'] + libs,
                             use = 'RobotDARTSimu RobotDARTMagnum',
-                            defines = ['RESPATH="' + path + '"'],
+                            defines = ['RESPATH="' + path + '"', 'GRAPHIC'],
```
costashatz, line 332: @jbmouret nice catch!

costashatz:
**Merging..**

Issue #63, opened 2020-05-18
Simplify APIs


jbmouret:
I feel there is some boilerplate code that can be deleted by exposing the objects directly. For instance, in graphics.hpp
```c++

 void set_speed(const Magnum::Vector2& speed) { _magnum_app->camera().set_speed(speed); }
                void set_near_plane(double near_plane) { _magnum_app->camera().set_near_plane(near_plane); }
                void set_far_plane(double far_plane) { _magnum_app->camera().set_far_plane(far_plane); }
                void set_fov(double fov) { _magnum_app->camera().set_fov(fov); }
                void set_camera_params(double near_plane, double far_plane, double fov, size_t width, size_t height) { _magnum_app->camera().set_camera_params(near_plane, far_plane, fov, width, height); }

                Magnum::Vector2 speed() const { return _magnum_app->camera().speed(); }
                double near_plane() const { return _magnum_app->camera().near_plane(); }
                double far_plane() const { return _magnum_app->camera().far_plane(); }
                double fov() const { return _magnum_app->camera().fov(); }

```
=> this can easily be replaced by a camera() accessor (then we can call fov, far_plane(), etc on the camera)

jbmouret:
- What is the convention: should we return a `gs::Camera&` or a gs::Camera*` ? I see both in the code (for instance, a pointer to the magnum_app but a reference to the lights)

- should we do the same for magnum_app?

```
  void set_recording(bool recording, bool recording_depth = false) { _magnum_app->record(recording, recording_depth); }
                bool recording() { return _magnum_app->recording(); }
                bool recording_depth() { return _magnum_app->recording_depth(); }

                bool shadowed() const { return _magnum_app->shadowed(); }
                bool transparent_shadows() const { return _magnum_app->transparent_shadows(); }
                void enable_shadows(bool enable = true, bool transparent = true) { _magnum_app->enable_shadows(enable, transparent); }
```

... or you want to hide this app?

costashatz:
> * What is the convention: should we return a `gs::Camera&` or a `gs::Camera*` ? I see both in the code (for instance, a pointer to the magnum_app but a reference to the lights)

This is true. We should only keep one of them. I would say that I prefer references. A small care should be taken with references and the python bindings, but I can be checking this.

> * should we do the same for magnum_app?

[...]

PR #66, opened 2020-05-19
try to simplify the graphics api (#63)

jbmouret:


costashatz:
@jbmouret I took the liberty to improve a bit the representation and speed of images and conversions. Now it is much faster and cleaner. I also remove duplicate code from `CameraOSR`. For the `BaseApplication`, I think we should leave it as is because it'll be too much work to change it now: I should do it at some point but another time. For now, I think it only remains to remove the inconsistencies between returning pointers and references.

costashatz:
**Merging..**

PR #70, opened 2020-05-25
Dof map

dalinel:
Adding dof name map index + mimic joint handling in robot.cpp


src/examples/talos.cpp
```diff
@@ -0,0 +1,79 @@
+#include <algorithm>
```
costashatz, line 1: We should either remove completely the `talos` example or make it self-containable, that is, include the meshes within `robot_dart`.

src/examples/talos.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+    simu.run(5.);
+    std::vector<std::string> dof_to_control;
+    dof_to_control.push_back("arm_left_4_joint");
+    dof_to_control.push_back("torso_2_joint");
+    Eigen::VectorXd cmd(2);
+    cmd(0) = 0.1;
+    cmd(1) = 0.1;
+
+    global_robot->set_actuator_type(global_robot->dof_index("arm_left_5_joint"), dart::dynamics::detail::ActuatorType::MIMIC, true);
```
costashatz, line 65: It might be a good idea to automatically update the maps once we call a function that might alter them. Of course we should still keep an `update_dof_map()` function for a user that might want to do something different/complicated.

src/robot_dart/robot.hpp
```diff
@@ -169,11 +177,21 @@ namespace robot_dart {
...
+        void _set_dof_data(int content, const Eigen::VectorXd& data, const std::vector<std::string>& dof_names = {}, bool filter_mimics = false, size_t start_dof = 0);
+        Eigen::VectorXd _get_vector_mimic(size_t start_dof, const Eigen::VectorXd& vec) const;
+        Eigen::VectorXd _set_vector_mimic(size_t start_dof, const Eigen::VectorXd& vec) const;
+
         std::string _robot_name;
         dart::dynamics::SkeletonPtr _skeleton;
         std::vector<RobotDamage> _damages;
         std::vector<std::shared_ptr<control::RobotControl>> _controllers;
-
+        std::unordered_map<std::string, size_t> _controllable_dof_map;
```
costashatz, line 189: Why do we need 3 different maps? One for controllable, one for full, and one for mimic joints? I would only have one map. What is the additional value that they are adding? What would be a use-case for this?

src/robot_dart/robot.hpp
```diff
@@ -97,17 +98,17 @@ namespace robot_dart {
         Eigen::Vector6d com_velocity() const;
         Eigen::Vector6d com_acceleration() const;
 
-        Eigen::VectorXd positions() const;
-        void set_positions(const Eigen::VectorXd& positions);
+        Eigen::VectorXd positions(const std::vector<std::string>& dof_names = {}, bool filter_mimics = false, size_t start_dof = 0);
```
costashatz, line 101: Why do we need the `start_dof` parameter in those functions? Same applies to the `update()` method..

Also, why do we need the `filter_mimics` parameter?

src/robot_dart/robot.hpp
```diff
@@ -169,11 +177,21 @@ namespace robot_dart {
         void _set_color_mode(dart::dynamics::MeshShape::ColorMode color_mode, dart::dynamics::SkeletonPtr skel);
         void _set_color_mode(dart::dynamics::MeshShape::ColorMode color_mode, dart::dynamics::ShapeNode* sn);
 
+        Eigen::VectorXd _get_dof_data(int content, const std::vector<std::string>& dof_names = {}, bool filter_mimics = false, size_t start_dof = 0);
```
costashatz, line 180: For performance, I will make these functions templated..

src/robot_dart/robot.hpp
```diff
@@ -169,11 +177,21 @@ namespace robot_dart {
...
+        void _set_dof_data(int content, const Eigen::VectorXd& data, const std::vector<std::string>& dof_names = {}, bool filter_mimics = false, size_t start_dof = 0);
+        Eigen::VectorXd _get_vector_mimic(size_t start_dof, const Eigen::VectorXd& vec) const;
```
costashatz, line 182: Why do we need the `start_dof` parameter?

src/robot_dart/robot_dart_simu.cpp
```diff
@@ -145,6 +145,29 @@ namespace robot_dart {
         _old_index = index;
     }
 
+    void RobotDARTSimu::step_once()
```
costashatz, line 148: For removing duplicate code, I will use this function inside the `run()` function.

src/examples/talos.cpp
dalinel: Maybe we can have a submodule pointing to [talos_robot](https://github.com/pal-robotics/talos_robot) ?

src/examples/talos.cpp
dalinel: Ok I will add this

src/robot_dart/robot.hpp
dalinel: Because I was creating the _controllable_dof_map it  was easy to create the _mimic_dof_map
And maybe the user will be interested to query those maps in an application

costashatz: My question was why do we need the `_controllable_map`, and why the `_full_dof_map` is not enough? Since this is a mapping from `names` to `indices`, why would we need different maps for the controllable joints and for the rest of the joints?

Maybe we need a function to get the names of the controllable/mimic/all dofs and not different maps?

src/robot_dart/robot.hpp
dalinel: The start_dof was initially used in robot_control.cpp it is an offset on the command
The filter_mimics boolean activate the functionality which was here in robot_control.cpp : 
You can send a cmd vector which do not contain mimic vector zero commands, those will be automatically added

I don't know if we really want that, but because we want to remove the control part of robot_dart, I have added it here

src/robot_dart/robot.hpp
dalinel: ok I will do that

src/robot_dart/robot.hpp
dalinel: It is what was on robot_control.cpp : an offset 
The user can send an cmd without the floating base for example

src/examples/talos.cpp
costashatz: I suggest to remove the talos example for now, as we will add a few robots to be officially supported by robot_dart in the future. Can you make an example with the KUKA arm for the new functionalities?

src/robot_dart/robot.hpp
costashatz: > The start_dof was initially used in robot_control.cpp it is an offset on the command

Yes, but start_dof was automatically identified in robot_control.

> You can send a cmd vector which do not contain mimic vector zero commands, those will be automatically added

But isn't `const std::vector<std::string>& dof_names` already doing that? You specify which dofs/joints to control with this. If you do not mention the mimic joints, they should receive zero.

> I don't know if we really want that, but because we want to remove the control part of robot_dart, I have added it here

I will handle this..

src/robot_dart/robot.hpp
costashatz: This is the same as the previous comment..

src/robot_dart/robot.hpp
dalinel: Yes this is a way for the user to get the names and index of the controllable/mimic/all dofs but we can also have a function to get the names only (instead of the different maps)

src/robot_dart/robot.hpp
dalinel: > But isn't const std::vector<std::string>& dof_names already doing that? You specify which dofs/joints to control with this. If you do not mention the mimic joints, they should receive zero.

Yes exactly, I have just added the functions of robot_control.cpp because as mentioned [here](https://github.com/resibots/robot_dart/issues/42#issuecomment-628015685), it is not logical to handle dof order both in robot.cpp and robot_control.cpp

> I will handle this..

ok

dalinel:
So I will add the KUKA arm example and the automatic update_dof_map() call after your modifications ? 
I can't write the example without knowing the exact content of the modifications that you will do

costashatz:
> So I will add the KUKA arm example and the automatic update_dof_map() call after your modifications ?

@dalinel I just finished my first pass of modifications. You can add the example and the automatic update dof map.

dalinel:
@costashatz I have added the example and added some `update_joint_dof_maps()` call

I have delete the `update_joint_dof_maps()` call in the constructors because now it is also called in `Robot::_set_damages(const std::vector<RobotDamage>& damages)`

For readability we can add it again in the constructor but I don't think it is needed

costashatz:
> @costashatz I have added the example and added some `update_joint_dof_maps()` call

Thanks! I will check them later today.

> I have delete the `update_joint_dof_maps()` call in the constructors because now it is also called in `Robot::_set_damages(const std::vector<RobotDamage>& damages)`
> 
> For readability we can add it again in the constructor but I don't think it is needed

Thanks for letting me know. I will check.

costashatz:
@dalinel thank you for this.. **Merging..**

PR #72, opened 2020-05-29
Clean-up Control

costashatz:
This PR adapts the control API to the new robot API (see #70) and cleans up a bit the namespace. This is breaking the API as I replaced all `std::vector`s with `Eigen::VectorXd`.


costashatz:
@jbmouret Cleaning up the control API, I actually believe that it is hard to make it much thinner. I do not really see how someone cannot fit their controller in this structure (the only downside with automated calls is that you need to pass objects around). Now that we have the map for controlling the joints, it is even easier to use. The only thing that could be cleaner would be to remove the multiple controllers and leave just one. But I do not see how this will make things better. Any ideas on what we could "cut down" to possibly fit your needs? Although we already have the step by step control thing.

jbmouret:
I think that if you want to keep a run(), then we need some API like this. I would have preferred to to have one (as I do not like the concept of 'run()'), but OK for me. I will add an example of the step-by-step control soon
.

jbmouret:
How do we handle the control mode (torque / velocity / torque+PID.)? is it part of control?

costashatz:
> How do we handle the control mode (torque / velocity / torque+PID.)?

We don't yet.

> is it part of control?

Not it is not yet.. I am trying to think about how to do it. I think that this should be a feature of the robot and then if a controller requires a specific control mode, it should assert it.

costashatz:
**Merging..***

Issue #52, opened 2020-05-12
Install the robot URDF in some 'path'?


jbmouret:
Should we install the URDF files in something like $PREFIX/robot_dart/robots/ ?

Because for now we need to copy/link them in each project.

jbmouret:
How do we include the URDF?
- as submodules?
- as files directly (with meshes, etc.)

And where do we install them?

I feel we need to have well-curated URDF because 'random urdf' usually do not work well. We also need to tune the collision shapes, etc. for maximum performance.

costashatz:
> I feel we need to have well-curated URDF

I agree with you here.

> And where do we install them?

Let's first finish the main restructuring of the library and then we should discuss on how to add example robots in robot_dart (along with their URDFs).

jbmouret:
We are almost there :) Thinking about it, I would suggest to create classes for each robot (Talos, Iiwa, etc.) because this would make it easier to define sensors and accessors with intuitive names.

Each class would use a default URDF, which would be taken from either a 'res' directory if it exists, from a ROBOT_DART_PATH env variable, or from `$PREFIX/share/robot_dart/robots` (installed with robot_dart). Of course, the user could use a different URDF with the same robot class (in that case, the user provides the path). $PREFIX should be taken from the compilation line.

The `res` directory is to make sure that the developers of robot_dart are using the urdf of the source tree and not the urdf that are installed on the system. The users should use the system files. The env. variable is to make it possible to have several installations of robot_dart files / switch etc.

PR #74, opened 2020-06-02
Install the URDF files in a system directory and search

jbmouret:
Currently only the search (WIP). See #52


jbmouret:
Question: do we need to use the same path prefix for the packages? how does that work?

jbmouret:
I broke the test_robot!

```
 /Users/jmouret/git/resibots/robot_dart/build/src/tests/test_robot

[...]
RobotDART:: using: /Users/jmouret/git/resibots/robot_dart/res/robots/pendulum.urdf
../src/tests/test_robot.cpp:58: error: in "test_fix_free": check pendulum->fixed() has failed
../src/tests/test_robot.cpp:60: error: in "test_fix_free": check !pendulum->free() has failed
unknown location:0: fatal error: in "test_fix_free": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000000)
../src/tests/test_robot.cpp:60: last checkpoint

*** 3 failures are detected in the test module "test_robot"
```
=> apparently the pointer has a problem, but I do not see where it comes from (yet). I guess the file is not properly loaded?

jbmouret:
It seems that the pendulum is not fixed by default. It works better if I first fix it with `pendulum->fix_to_world();` but it then segv when we call `pendulum->free_from_world();`. Sounds like a bug somewhere but this is unexpected.


`

jbmouret:
@costashatz What do we do for complex URDF (Talos, iCub, etc.)?
- option 1: they have they urdf in their own repo
- option 2: we store in this repo a good urdf for simulation
- option 3: we use git submodule to insert the repo of the urdf

It might be interesting to use a repo for these complex robots because the urdf are also used in the control libraries (tsid, opensot, etc.). However:
- having multiple repos is not very user-friendly (makes things a bit more complicated)
- I suggest to have a Talos class, but this would mean that this class can only work if we have cloned the right urdf repo (=> the code would not work out of the box)
- We might not use the same urdf for control and for simulation anyway
- I don't like having many repos without some directory structure (which is not given by github...) to find things that work together

I find it interesting that you had to use a special URDF from talos (not exactly the one we used so far): this suggests that having a special URDF for simulation is needed, and suggests that we should have our own urdf library. Of course, the modified URDF might work for control too (and Gazebo). But is it the rule? the same urdf for dart, gazebo, tsid, openSOT? is it easy?

I guess your are leaning towards option 3 (using a submodule)? I would prefer option 1 but no option is perfect.

What do you think?

jbmouret:
There is also option 4: 
a urdf library repo that contains good urdfs for all the robots (all the urdf at the same place, but outside of robot_dart). But, as we said, either we use this as a submodule, or we cannot really create class for robots (as they would be meaningless without a particular urdf).

So... option 1: we simply store all the urdf in robot_dart?

Aneoshun:
With option 1, how do you plan to deal with situations such as we have different hexapod URDFs, because my new hexapod is slightly different from the Inria hexapod. 
Obviously, I want to use the URDF of my robot, not the Inria one, but storing all the URDFs of all the variants of the robots is not realistic.

jbmouret:
@costashatz I added a Talos example. Questions:
- It seems that the packages require absolute paths? how does that work?
- I've put a URDF and meshes for Talos that I had, but they might not be your favorite version (please check or replace by yours)
- Should we keep the xacro here? I guess they are useful if we need to regenerate some things? (I did not do this for a long long time...).

jbmouret:
@Aneoshun if you your hexapod urdf, can you send it to me?

costashatz:
> * It seems that the packages require absolute paths? how does that work?

If I remember well yes, they require absolute paths.

> * I've put a URDF and meshes for Talos that I had, but they might not be your favorite version (please check or replace by yours)

I will..

> * Should we keep the xacro here? I guess they are useful if we need to regenerate some things? (I did not do this for a long long time...).

I would say in principle no, but it might be a good idea to keep the xacros used to generate the URDF. This can be good for debugging or small improvements; easier also for other people to improve the files.

jbmouret:
Another question: what is the 'is_urdf_string' boolean?

costashatz:
> Another question: what is the 'is_urdf_string' boolean?

One can pass a string containing the contents of a URDF file (instead of reading the file). This is useful when reading the URDF from a parameter server (see ROS) or similar cases.

jbmouret:
Thanks. This makes it hard to know where to find the package: currently I am searching for the main urdf file to know if we use a absolute path / relative path / env variable  / system path.
... this means that I cannot do this (since there is no filename) and I have to actually search for the packages too.

jbmouret:
Question (@costashatz) : I would like that:
- when we are in the robot_dart directory, we use the robot_dart urdf files
- otherwise, we use the files installed in the system

=> this is straightforward to do if I remove the 'res' and move the 'robots' directory at the root of the robot_dart directory. To me it makes sense. Did you have any other 'ressources' in mind?

costashatz:
> To me it makes sense.

To me as well. Please do it. The robots directory only contains meshes and URDF files, right?

> Did you have any other 'ressources' in mind?

Not really. At least for the moment.

jbmouret:
@costashatz Could you start having a look at this? I think I made most of the changes.

Next step will be to create classes for the robots.

src/examples/pendulum.cpp
```diff
@@ -37,7 +37,7 @@ int main()
     // dmg.data = "arm_joint_3";
     // brk.push_back(dmg);
 
-    auto global_robot = std::make_shared<robot_dart::Robot>("res/models/pendulum.urdf");
+    auto global_robot = std::make_shared<robot_dart::robots::Robot>("pendulum.urdf");
```
costashatz, line 40: Why did you move `Robot.hpp` to a different namespace and folder?

wscript
```diff
@@ -172,11 +175,12 @@ def build(bld):
     robot_dart_magnum_srcs = " ".join(magnum_files)
 
     libs = 'BOOST EIGEN DART PTHREAD'
-
+    defines = ["ROBOT_DART_PREFIX=\"" + bld.env['PREFIX'] + "\""]
```
costashatz, line 178: We need to create a `config.hpp` that gets installed with the library and contains these directories for people trying to use the installed library. Even if they do not really need it, it's good have them automatically generated and saved somewhere for reference.

src/examples/pendulum.cpp
jbmouret: I wanted to make space for having the robot-specific classes (which in my mind was going with having the URDF organized). However, we should probably make a second PR for this.

wscript
jbmouret: OK. I can do this.

src/examples/pendulum.cpp
jbmouret: (by the way, git does not track the 'git mv' and therefore this creates conflict with any change to the Robot class in master... I did not expect this).

costashatz: Even if we have robot-specific classes, why would we need to move the base `Robot` class? I am okay with it, but it will break compatibility too much. I would keep the `Robot` class in the `robot_dart` namespace and move all the robot-specific classes in `robot_dart::robots` namespace.

jbmouret: I wanted to follow your convention of having the abstract class in the directory/namespace of the concrete classes. You did this for control, for instance (robot_control.hpp). And we want to haves all the robots in a `robots` namespace.

costashatz: I would say that in this case `Robot` is not an abstract class, but a main ingredient of `robot_dart`. But in any case this is not a very big deal, I can move it before merging back to the main namespace.

jbmouret: I can move it back if you prefer. It will make merging easier (btw, what do you use for indenting? I will need to reindent this file and it seems that my vs code does not do the same as yours)

costashatz: > I can move it back if you prefer.

Please do if you wish.

> btw, what do you use for indenting? I will need to reindent this file and it seems that my vs code does not do the same as yours

I am using clang-format and this format file: https://github.com/resibots/format_code/blob/master/.clang-format (just copy-paste it in your home folder).

wscript
```diff
@@ -6,7 +6,7 @@ import fnmatch
 import glob
 sys.path.insert(0, sys.path[0]+'/waf_tools')
 
-VERSION = '1.0.0'
+VERSION = '1.0'
```
costashatz, line 9: Please keep the patch version also. We might need it in the future and it does not hurt anywhere. Users can still query `v>=1.1` (i.e., ignoring the patch version).

jbmouret: OK. I never use the patch version :) Do I do like MAJOR/MINOR/PATCH ?

jbmouret:
It seems that the test_control.cpp is broken:
`robot_dart assertion failed: RobotControl: parent robot should be initialized; use set_robot()` 
(many times)

costashatz:
> It seems that the test_control.cpp is broken:
> `robot_dart assertion failed: RobotControl: parent robot should be initialized; use set_robot()`
> (many times)

I saw it and I'll fix it. No worries.

jbmouret:
I've added iCub but I do not get the materials right (it's all blue). @costashatz Any idea?

PedroDesRobots:
Yes, we were using` talos_full_v2.urd.xacro` with parameters set as

-  `enable_crane := false` 

-  `foot_collision := thinbox `

STL files are lighter but they are not encoded color or texture..  So the result give a gray Talos. It should be possible to add colors in the post process (after generating the urdf).

costashatz:
> STL files are lighter but they are not encoded color or texture

For Talos we do not really need texture. It's black/white/orange which is easy to encode through the URDF.

I'll try to replace all the meshes with `GLFT` or `PLY` which are binary and much better than `STL`. DART (assimp) can still handle them (in theory).

jbmouret:
So, what is the status for Talos? How can we help?

costashatz:
> So, what is the status for Talos?

I tested the usage of `gltf2` meshes and it is working and they are half the size.

> How can we help?

You can help if you have blender and convert all the dae meshes to `glft2` (binary format). Make sure to export only the selection (select the mesh) and do not export materials (as we do not really need them for Talos).

I will do this until the end of the week, but you can be faster I guess..

jbmouret:
OK. If we need to select, then it cannot be scripted?

costashatz:
> OK. If we need to select, then it cannot be scripted?

It should be able to be scripted as the file only has the mesh. I said to select as the blender scene has lights etc, but if you know how to use the python-blender interface it should be doable via scripts.

jbmouret:
This is supposed to do what we want:
```python
import bpy
import os

path = os.getcwd() + '/talos_description/meshes/'  # set this path
print("PATH:", path)

for root, dirs, files in os.walk(path):
    for f in files:
        print("f:", f, root)
        if f.endswith('.dae') :
            dae_file = os.path.join(root, f)
            gltf_file = os.path.splitext(dae_file)[0]
            print(dae_file, "=>", gltf_file)

            bpy.ops.object.select_all(action='SELECT')
            bpy.ops.object.delete()
            bpy.ops.wm.collada_import(filepath=dae_file)
            bpy.ops.object.select_all(action='SELECT')
            bpy.ops.export_scene.gltf(filepath=gltf_file, ui_tab='MESHES')
```

Run like:
```
blender --background --python blender.py
```

This runs without any error, apparently (it exports in `glb`, is it right?).

However, when I import a DAE or a GLB (or a STL) in the GUI or in the script, nothing appears in my Blender scene (and there is nothing in the three). So... how did you manage to import a DAE in blender? I have no error, nothing... 

By the way, maybe this could help us to fix the normals?

costashatz:
> it exports in `glb`, is it right?

Yes `glb` is the extension.

> If I remove the effect in the DAE, then I can load it in blender.
> 
> => our DAE would need manual editing (via a script...) to load in blender.

Is it for all DAEs? I remember that I loaded the torso without any problem.

> It it worth it? gltf would make things smaller and faster to load, but do we care that much?

We do care somehow: 1) since everything goes into the repo, we do not want the repo to be 500mb (it can easily get there if we start adding DAEs that change a bit over time etc.), 2) if it loads faster, it gives better experience to the user. But it's not the most important issue of the repo/library.

> I would say if I can fix the normals, maybe it's worth it.

If we can fix the normals also, then it is even more worth it.

jbmouret:
The collision DAEs work fine (they probably have been simplified). Some objects might not have the 'effect' included.

Btw, the difference is not that huge: 
- arm_5.dae: 2.1 M
- arm_5.glb: 1.2 M

I agree that a smaller repo and faster loading is good, but DAE is also more widely supported (e.g., meshlab does not understand glb files).

I tried to smooth the normals but no success so far: when ask blender to smooth the STL, it works; but the same operation on a DAE or GLB file does nothing.

jbmouret:
More info. I managed to smooth the DAE by first using meshlab to merge close vertices (with distance=0). Apparently, each face is treated as independent, which prevents from computing vertex normals.

Incidentally, doing this reduces the size of the DAE... to the size of of the glb!
- arm_5-meshlab.dae : 1.3 M

... but fixing the normals in meshlab then export the normals does not work in robot_dart. However, importing the DAE, smoothing in blender, then export to glb... gives us the right normals!

I will try to script all of this and see...

jbmouret:
I got a nice URDF with GLB meshes and smoothing!

However, the glb and DAE are about the same size once fixed by meshlab (13 M for DAE and 15M for glb). If we target the size, this is wrong. However, we do get a smooth rendering with gltf, which is definitely a good thing.

@costashatz how do you want to proceed? should I commit? make a branch? maybe a git push is not the best since we might want to iterate a bit.

Also, there is the texture that is left to remove (and replace it with a color). I don't know what is best to do this.

In addition, for the Talos URDF:
- we should make the legs less black (so that we see them)
- we can optimize further the collision meshes
- the shadows still do these visual artefacts

costashatz:
> However, the glb and DAE are about the same size once fixed by meshlab (13 M for DAE and 15M for glb)

This is weird. `dae` is xml and `glb` is binary and they are storing the same information; according to their specifications, `glb` stores even less data. So this is not possible technically. I guess `meshlab` or `blender` are saving something weirdly (or some command is missing some argument).

> Also, there is the texture that is left to remove (and replace it with a color). I don't know what is best to do this.

Remove completely the texture, and just put colors: white, black (you can make this dart gray) and orange.

> * we can optimize further the collision meshes

I would do simple meshes (boxes, cylinders, spheres). Apart from being much faster to compute the collision checking, it makes it far more robust for the collision checker (as many triangle meshes are not well-defined; Talos' seem also to not be perfectly defined).

> * the shadows still do these visual artefacts

The shadows have nothing to do with Talos specifically. It is just that the meshes of Talos have (1) the weird scaling (negative sign) and (2) many bumps within the same mesh that a small shadow map (like the one I am using cannot handle). Try creating a graphics config for the visualizer where the [shadow map size](https://github.com/resibots/robot_dart/blob/master/src/robot_dart/gui/magnum/base_application.hpp#L95) is 2048 (see [here](https://github.com/resibots/robot_dart/blob/master/src/examples/cameras.cpp#L44-L47) for an example on how to set the graphics config).

> Without the shadows:
> Screenshot 2020-06-17 at 13 03 22

It seems that you have removed a lot of triangles from the mesh. Can't we have access to the CAD files from PAL? Or a very high-detailed mesh? Then, we can generate nice meshes (either `dae` or `glb`) with fewer triangles and nice normals.

> @costashatz how do you want to proceed? should I commit? make a branch? maybe a git push is not the best since we might want to iterate a bit.

Can you first send me the `glb` files? I'll try to see why they are bigger than `dae`. Can you also send me the scripts that you used?

jbmouret:
> > However, the glb and DAE are about the same size once fixed by meshlab (13 M for DAE and 15M for glb)
> 
> This is weird. `dae` is xml and `glb` is binary and they are storing the same information; according to their specifications, `glb` stores even less data. So this is not possible technically. I guess `meshlab` or `blender` are saving something weirdly (or some command is missing some argument).

I did not expect this either...


> > Also, there is the texture that is left to remove (and replace it with a color). I don't know what is best to do this.
> 
> Remove completely the texture, and just put colors: white, black (you can make this dart gray) and orange.

The question is how to script this properly. I think I can set the color with blender but I have to investigate.

> > * we can optimize further the collision meshes
> 
> I would do simple meshes (boxes, cylinders, spheres). Apart from being much faster to compute the collision checking, it makes it far more robust for the collision checker (as many triangle meshes are not well-defined; Talos' seem also to not be perfectly defined).

Yes, this would be ideal, but not trivial to script because of the rotations (hard to guess from the mesh how the part is rotated). I guess the only way is to do it by hand...
 
> It seems that you have removed a lot of triangles from the mesh. Can't we have access to the CAD files from PAL? Or a very high-detailed mesh? Then, we can generate nice meshes (either `dae` or `glb`) with fewer triangles and nice normals.

These DAE are from PAL. I agree that they are already decimated (I don't understand why). The STL look the same to me... I we can try to get better meshes, but I do not expect this to be easy/fast.

> 
> > @costashatz how do you want to proceed? should I commit? make a branch? maybe a git push is not the best since we might want to iterate a bit.
> 
> Can you first send me the `glb` files? I'll try to see why they are bigger than `dae`. Can you also send me the scripts that you used?

https://filesender.renater.fr/?s=download&token=b82a42db-620d-4d2f-b4ac-c0d3f19f0a55

From the talos directory:
- You first need to run the meshlab script (which requires meshlabserver in your path), which creates `basename`-meshlab.dae
- Then the `blender --background --python blender2.py`

jbmouret:
It is nicer with 2048 for the body, still not perfect for the forearm and the feet. What is the scaling issue exactly? while we are at fixing meshes maybe we can try to fix this too.
Screenshot 2020-06-17 at 22 13 18

jbmouret:
Btw, I think gltf only handles meshes, not sphere or cylinders. Therefore there is no point in using anything else than boxes for collision if we use gltf. 

I think Collada might be able to define spheres and cylinders, but will Dart work out of the box with this (without converting to meshes)?

costashatz:
> Yes, this would be ideal, but not trivial to script because of the rotations (hard to guess from the mesh how the part is rotated). I guess the only way is to do it by hand...

Yes, by hand is by far the easiest/safest way. Also keep in mind that we do not need collision shapes for every visual shape necessarily.

> These DAE are from PAL. I agree that they are already decimated (I don't understand why). The STL look the same to me... I we can try to get better meshes, but I do not expect this to be easy/fast.

Even if we do not have the meshes fast, I think it'd be worth it of having some highly detailed meshes for Talos in the long run, so that we can use them to create our visualization meshes.

> still not perfect for the forearm and the feet

This is because of the negative scaling.

> What is the scaling issue exactly? while we are at fixing meshes maybe we can try to fix this too. (is it when they use -1 to do a symmetry?)

Yes this is why this happens. In theory, these artifacts should not happen, but no matter what I tried in code  they still happen. Give me today to try to fix this in code. If it doesn't work, we should import the meshes in blender and properly mirror them (we might need to recalculate the normals).

> Btw, I think gltf only handles meshes, not sphere or cylinders. Therefore there is no point in using anything else than boxes for collision if we use gltf.
> 
> I think Collada might be able to define spheres and cylinders, but will Dart work out of the box with this (without converting to meshes)?

I was thinking of directly writing this in the URDF (using the URDF commands for basic meshes), that is, no meshes at all. If we do not do it that way, DART will always handle them as full meshes.

> I get 12M for the `glb` files by removing the materials (including the texture?) with blender

I am not familiar with the python scripting for Blender at all. We should not need the textures; they are just colors, not really a texture.

jbmouret:
> I am not familiar with the python scripting for Blender at all. We should not need the textures; they are just colors, not really a texture.

Yes, it works fine when I do this.

It should be easy to mirror the meshes in blender.

costashatz:
> Give me today to try to fix this in code.

I found a way to fix this, but it requires a small change in the gui lib. I'll try to push this today..

costashatz:
> I found a way to fix this, but it requires a small change in the gui lib. I'll try to push this today..

Found an easier fix. This is what Talos looks like with shadow map size of 2048:

image

The fix makes the shadows a bit more pixelish (if not used with a big shadow map size):

image

But it's not that bad, and someone can increase the shadow map size (or we can make the shadow map size bigger by default --- small laptops are going to have a hard time though!).

jbmouret:
I think this is good!

costashatz:
@jbmouret can you verify that now the shadows do not have artifacts? and that the Talos `black` is now `dart-grey`?

jbmouret:
Yes, the shadows are much better now! The Talos is indeed in grey instead of black (which is nice). While we are at it, we should try to reduce the 'specular' component of the lights (is in the material or in magnum?).

costashatz:
> While we are at it, we should try to reduce the 'specular' component of the lights (is in the material or in magnum?).

This is me and DART. It's DART because they do not load/use any specular material, and it is me because I set as a default one with very big values. I can change this easily.

jbmouret:
Now that we have finished these proposals. What is left before merging? we wanted to improve Talos's urdf, but maybe we can do this in a different PR?

costashatz:
> What is left before merging?

I think nothing is left. I will try to do as soon as possible a last check in the code.

> we wanted to improve Talos's urdf, but maybe we can do this in a different PR?

Yes. This should go to a different PR.

> we should try to reduce the 'specular' component of the lights

Let's fix this also in a different PR. Can you make an issue about this? And assign me..

jbmouret:
I tried to remove the specular effect (white spots) and I played with the default 'shininess'. That did not work. The only thing that makes things better is:
```
            highp float specularity = 0.1 * pow(max(0.0, dot(normalize(cameraDirection), reflection)), shininess + 1e-8);
```

(I added a 0.1 *) in `PhongMultiLight.frag`.

costashatz:
> [@jbmouret ](/jbmouret)
> ` [our Talos has red grippers!](/resibots/robot_dart/pull/74/commits/9638cc12fe84f2478e473d64853f5497114c0e3d)`

You are using the `glb` meshes but you haven't pushed the meshes..  Please avoid merging like this.. We should handle this in a different PR. Can you undo your commit? Or let me undo it?

> I tried to remove the specular effect (white spots) and I played with the default 'shininess'. That did not work.

I'll fix this. Do not worry about this. I'll do it in another PR..

jbmouret:
Sorry, this was truly not voluntary... I don't know what happened. I actually thought the glb were already pushed. Please undo what you want.

costashatz:
@jbmouret I had a deeper look, and :

- I cleaned the `dae` files of Talos
- There are a few `stl` ones remaining for Talos
- We need to clean the iCub files also

I'll do this as soon as possible.. After this, I think that we can merge..

costashatz:
@jbmouret can you make the visual meshes of iCub (`dae` fiels) to `glb`?

jbmouret:
I will have a look. This sounds possible.

jbmouret:
I had to rotate the visual of the feet (I don't know why). Also, I am not fond of the color, but this seems to be embedded in the glb (not in the urdf).

jbmouret:
> * There are a few `stl` ones remaining for Talos

Why can't we remove them? do we use them?

costashatz:
> Why can't we remove them? do we use them?

I will remove them. I just have to do it one by one (because there are other `stl` that we use)...

costashatz:
> I had to rotate the visual of the feet (I don't know why).

I guess we have to rotate the waist also.. I'll try to do it..

> I am not fond of the color

You mean of the feet or other color? The rest seems the same to me..

costashatz:
> Also, I am not fond of the color, but this seems to be embedded in the glb (not in the urdf).

The issue is the normals, not the color. They are different from the dae files. Can you tell me what you did? And I can try to do it myself?

costashatz:
@jbmouret now that I check the Talos meshes also, the `glb` meshes look a bit weird to me. Like much less triangles than the `dae` ones. Did you do any transformation of the meshes?

jbmouret:
I did not remove any triangle but I smoothed the normals. Smoothing is done using a heuristics (< 12 degrees), so sometimes we lose details (the triangles are still there, but they are too smooth). I simply did:

```python
 for obj in bpy.data.objects:
                if obj.type == 'MESH':
                    obj.data.use_auto_smooth = 1
                    obj.data.auto_smooth_angle = math.radians(5) # 12 degrees
```


For iCub, I do not see any difference with the DAE version except for the feet (which are smoother now) and the waist (as you noted). I do not like the green/blue color, but we will see this kind of detail later :)

costashatz:
> For iCub, I do not see any difference with the DAE version except for the feet

For me the shading is different (I am assuming because of the normals), but seems a bit weird.

For Talos, the meshes seem a bit too smoothed.

> Smoothing is done using a heuristics (< 12 degrees)

Can you try with smaller degrees?

costashatz:
@jbmouret I think that this is ready to be merged. Can you have a last look?

costashatz:
**Merging...**

Issue #75, opened 2020-06-03
test_robot.cpp fails?


jbmouret:
In the master:
```c++

 /Users/jmouret/git/resibots/robot_dart/build/src/tests/test_robot

Running 4 test cases...
Entering test module "test_robot"
../src/tests/test_robot.cpp:16: Entering test case "test_constructors"
Warning [LocalResource.cpp:48] [LocalResource::constructor] Failed opening file '/tmp/bad-path.urdf' for reading: No such file or directory
Warning [CompositeResourceRetriever.cpp:97] [CompositeResourceRetriever::retrieve] All ResourceRetrievers registered for this schema failed to retrieve the URI 'file:///tmp/bad-path.urdf' (tried 1).
../src/tests/test_robot.cpp:19: info: check 'exception "Assertion" raised as expected: validation on the raised exception through predicate "[](const Assertion&) { return true; }"' has passed
../src/tests/test_robot.cpp:22: info: check 'exception "Assertion" raised as expected: validation on the raised exception through predicate "[](const Assertion&) { return true; }"' has passed
../src/tests/test_robot.cpp:31: info: check robot == nullptr has passed
../src/tests/test_robot.cpp:35: info: check pendulum has passed
../src/tests/test_robot.cpp:39: info: check dummy has passed
../src/tests/test_robot.cpp:42: info: check pendulum->name() == "robot" has passed
../src/tests/test_robot.cpp:43: info: check pendulum->name() == pendulum->skeleton()->getName() has passed
../src/tests/test_robot.cpp:45: info: check dummy->name() == "dummy_robot" has passed
../src/tests/test_robot.cpp:46: info: check dummy->name() == dummy->skeleton()->getName() has passed
../src/tests/test_robot.cpp:47: info: check dummy->name() == dummy_skel->getName() has passed
../src/tests/test_robot.cpp:16: Leaving test case "test_constructors"; testing time: 95039us
../src/tests/test_robot.cpp:52: Entering test case "test_fix_free"
../src/tests/test_robot.cpp:55: info: check pendulum has passed
../src/tests/test_robot.cpp:59: info: check pendulum->fixed() has passed
../src/tests/test_robot.cpp:61: info: check !pendulum->free() has passed
unknown location:0: fatal error: in "test_fix_free": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000000)
../src/tests/test_robot.cpp:61: last checkpoint
Test is aborted
[...]

PR #76, opened 2020-06-04
Fix Catalina bug

jbmouret:
see #75

### DIFF
--- a/src/python/example.py
+++ b/src/python/example.py
@@ -30,7 +30,7 @@ class MyDesc(rd.Descriptor):
             self._states.append(self._simu.robot(0).positions())
 
 # Load robot from URDF
-robot = rd.Robot("res/robots/arm.urdf", "arm", False)
+robot = rd.Robot("res/models/arm.urdf", "arm", False)
 robot.fix_to_world()
 
 # Initiate custom controller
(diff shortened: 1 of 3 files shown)


src/python/example.py
```diff
@@ -30,7 +30,7 @@ def __call__(self):
             self._states.append(self._simu.robot(0).positions())
 
 # Load robot from URDF
-robot = rd.Robot("res/models/arm.urdf", "arm", False)
+robot = rd.Robot("res/robots/arm.urdf", "arm", False)
```
costashatz, line 33: I guess these are left-overs from another branch right?

src/python/example_parallel.py
```diff
@@ -5,7 +5,7 @@
 from timeit import default_timer as timer
 import os
 
-robot = rd.Robot("res/models/arm.urdf", "arm", False)
+robot = rd.Robot("res/robots/arm.urdf", "arm", False)
```
costashatz, line 8: > I guess these are left-overs from another branch right?

Same here.

jbmouret: Yes. Sorry, I had a few things staged and I wrote 'commit' a bit too fast. This is from urdf_library...

costashatz: Can you revert them? (or just do a new commit that is undoing them)

costashatz:
**Merging..**

Issue #77, opened 2020-06-04
RobotDARTConfig.cmake.in requires eigen3 3.3


dalinel:
In [RobotDARTConfig.cmake.in:](https://github.com/resibots/robot_dart/blob/master/cmake/RobotDARTConfig.cmake.in) 
`find_dependency(Eigen3 3.3 REQUIRED NO_MODULE)`
Forces the user to have Eigen3 with a version > 3.3 which is not the case in the waf installation
Can it be changed to `find_dependency(Eigen3 REQUIRED NO_MODULE)` ?

costashatz:
DART requires Eigen v3.9.2 at least (see [here](https://github.com/dartsim/dart/blob/master/cmake/DARTFindEigen3.cmake#L13)), so I think we should adapt both CMake and `waf` to require this.

jbmouret:
You mean 3.2.92 right?

costashatz:
> You mean 3.2.92 right?

Yes. You are right..

PR #78, opened 2020-06-04
Fix eigen version (WIP)

jbmouret:
#77 

Currently check only in waf


jbmouret:
pending @costashatz's approval

PR #80, opened 2020-06-05
Actuator Types

costashatz:
This PR adds aliases for `actuator types` of DART which we call `control modes`. @jbmouret I was thinking which control modes to provide to the user and I ended up with the conclusion that we should not provide any position control (either kinematic or dynamics with torques) as this would interfere with the control API. The user can use the step by step API to make a PID controller (they need to add one line of code!) and with the new functions they can change the control modes per dof/joint in a DART-free way (only by strings).

### DIFF
--- a/cmake/example/example.cpp
+++ b/cmake/example/example.cpp
@@ -31,7 +31,7 @@ int main()
     // g_robot->skeleton()->setPosition(1, M_PI / 2.0);
     Eigen::Vector3d size(0, 0, 0);
 
-    global_robot->set_actuator_types(dart::dynamics::Joint::VELOCITY);
+    global_robot->set_actuator_types("velocity");
 
     Eigen::VectorXd ctrl(4);
     ctrl << 0.0, 1.0, -1.5, 1.0;
--- a/src/examples/arm.cpp
+++ b/src/examples/arm.cpp
@@ -31,7 +31,7 @@ int main()
     // g_robot->skeleton()->setPosition(1, M_PI / 2.0);
     Eigen::Vector3d size(0, 0, 0);
 
-    global_robot->set_actuator_type("velocity");
+    global_robot->set_actuator_types("velocity");
 
     Eigen::VectorXd ctrl(4);
     ctrl << 0.0, 1.0, -1.5, 1.0;
--- a/src/examples/arm.cpp
+++ b/src/examples/arm.cpp
@@ -31,7 +31,7 @@ int main()
     // g_robot->skeleton()->setPosition(1, M_PI / 2.0);
     Eigen::Vector3d size(0, 0, 0);
 
-    global_robot->set_actuator_types(dart::dynamics::Joint::VELOCITY);
+    global_robot->set_actuator_type("velocity");
 
     Eigen::VectorXd ctrl(4);
     ctrl << 0.0, 1.0, -1.5, 1.0;
--- a/src/examples/dof_maps.cpp
+++ b/src/examples/dof_maps.cpp
@@ -20,7 +20,7 @@ int main()
     global_robot->set_position_enforced(true);
 
     // Set actuator types to SERVO motors so that they stay in position without any controller
-    global_robot->set_actuator_type("servo");
+    global_robot->set_actuator_types("servo");
 
     float dt = 0.001;
     robot_dart::RobotDARTSimu simu(dt);
@@ -76,8 +76,8 @@ int main()
     bool filter_mimics = true, filter_locked = true, filter_passive = true;
 
     global_robot->set_mimic("iiwa_joint_5", "iiwa_joint_1");
-    global_robot->set_actuator_type("locked", {"iiwa_joint_6"});
-    global_robot->set_actuator_type("passive", {"iiwa_joint_7"});
+    global_robot->set_actuator_type("locked", "iiwa_joint_6");
+    global_robot->set_actuator_type("passive", "iiwa_joint_7");
 
     //Get the controllable joints (You cannot send a command to a mimic or passive or locked joint)
     auto controllable_dofs = global_robot->dof_names(filter_mimics, filter_locked, filter_passive);
--- a/src/examples/dof_maps.cpp
+++ b/src/examples/dof_maps.cpp
@@ -20,7 +20,7 @@ int main()
     global_robot->set_position_enforced(true);
 
     // Set actuator types to SERVO motors so that they stay in position without any controller
-    global_robot->set_actuator_types(dart::dynamics::Joint::SERVO);
+    global_robot->set_actuator_type("servo");
 
     float dt = 0.001;
     robot_dart::RobotDARTSimu simu(dt);
@@ -73,16 +73,11 @@ int main()
     // }
 
     /**************** Mimic / Passive / Locked joint handling ********************************************************/
-
-    int mimic_index = global_robot->dof_index("iiwa_joint_5");
-    int locked_index = global_robot->dof_index("iiwa_joint_6");
-    int passive_index = global_robot->dof_index("iiwa_joint_7");
-
     bool filter_mimics = true, filter_locked = true, filter_passive = true;
 
-    global_robot->set_actuator_type(mimic_index, dart::dynamics::Joint::MIMIC, true);
-    global_robot->set_actuator_type(locked_index, dart::dynamics::Joint::LOCKED, true);
-    global_robot->set_actuator_type(passive_index, dart::dynamics::Joint::PASSIVE, true);
+    global_robot->set_mimic("iiwa_joint_5", "iiwa_joint_1");
+    global_robot->set_actuator_type("locked", {"iiwa_joint_6"});
+    global_robot->set_actuator_type("passive", {"iiwa_joint_7"});
 
     //Get the controllable joints (You cannot send a command to a mimic or passive or locked joint)
     auto controllable_dofs = global_robot->dof_names(filter_mimics, filter_locked, filter_passive);
--- a/src/examples/hexapod.cpp
+++ b/src/examples/hexapod.cpp
@@ -12,7 +12,7 @@ int main()
 
     global_robot->set_position_enforced(true);
 
-    global_robot->set_actuator_type("servo");
+    global_robot->set_actuator_types("servo");
     global_robot->skeleton()->enableSelfCollisionCheck();
 
     auto g_robot = global_robot->clone();
--- a/src/examples/hexapod.cpp
+++ b/src/examples/hexapod.cpp
@@ -12,7 +12,7 @@ int main()
 
     global_robot->set_position_enforced(true);
 
-    global_robot->set_actuator_types(dart::dynamics::Joint::SERVO);
+    global_robot->set_actuator_type("servo");
     global_robot->skeleton()->enableSelfCollisionCheck();
 
     auto g_robot = global_robot->clone();
--- a/src/python/robot.cpp
+++ b/src/python/robot.cpp
@@ -46,9 +46,15 @@ namespace robot_dart {
                 .def("fixed", &Robot::fixed)
                 .def("free", &Robot::free)
 
-                .def("set_actuator_type", &Robot::set_actuator_type,
+                .def("set_actuator_types", &Robot::set_actuator_types,
                     py::arg("type"),
                     py::arg("joint_names") = std::vector<std::string>(),
+                    py::arg("override_mimic") = false,
+                    py::arg("override_base") = false)
+
+                .def("set_actuator_type", &Robot::set_actuator_type,
+                    py::arg("type"),
+                    py::arg("joint_name"),
                     py::arg("override_mimic") = false,
                     py::arg("override_base") = false)
 
--- a/src/python/robot.cpp
+++ b/src/python/robot.cpp
@@ -63,6 +63,23 @@ namespace robot_dart {
                 .def("actuator_type", &Robot::actuator_type)
                 .def("actuator_types", &Robot::actuator_types)
 
+                .def("set_control_mode", &Robot::set_control_mode,
+                    py::arg("mode"),
+                    py::arg("dof_names") = std::vector<std::string>(),
+                    py::arg("override_mimic") = false,
+                    py::arg("override_base") = false)
+
+                .def("set_mimic", &Robot::set_mimic,
+                    py::arg("dof_name"),
+                    py::arg("mimic_dof_name"),
+                    py::arg("multiplier") = 1.,
+                    py::arg("offset") = 0.)
+
+                .def("control_mode", &Robot::control_mode)
+
+                .def("control_modes", &Robot::control_modes,
+                    py::arg("dof_names") = std::vector<std::string>())
+
                 .def("set_position_enforced", (void (Robot::*)(size_t, bool)) & Robot::set_position_enforced)
                 .def("set_position_enforced", (void (Robot::*)(const std::vector<bool>&)) & Robot::set_position_enforced)
                 .def("set_position_enforced", (void (Robot::*)(bool)) & Robot::set_position_enforced)
--- a/src/python/robot.cpp
+++ b/src/python/robot.cpp
@@ -48,20 +48,20 @@ namespace robot_dart {
 
                 .def("set_actuator_type", &Robot::set_actuator_type,
                     py::arg("type"),
-                    py::arg("dof_names") = std::vector<std::string>(),
+                    py::arg("joint_names") = std::vector<std::string>(),
                     py::arg("override_mimic") = false,
                     py::arg("override_base") = false)
 
                 .def("set_mimic", &Robot::set_mimic,
-                    py::arg("dof_name"),
-                    py::arg("mimic_dof_name"),
+                    py::arg("joint_name"),
+                    py::arg("mimic_joint_name"),
                     py::arg("multiplier") = 1.,
                     py::arg("offset") = 0.)
 
                 .def("actuator_type", &Robot::actuator_type)
 
                 .def("actuator_types", &Robot::actuator_types,
-                    py::arg("dof_names") = std::vector<std::string>())
+                    py::arg("joint_names") = std::vector<std::string>())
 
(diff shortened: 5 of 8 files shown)


jbmouret:
Couldn't the position control "mode" be a controller that we add? I do not think this is "one line" to have a PID controller for each joint of the Talos (including that we would have different PID values for different joints).

Also, this mode could be in the Talos class.

My general point of view is that simple things should be straighforward: moving a robot in position control should be the most straightforward thing to do.

costashatz:
> Couldn't the position control "mode" be a controller that we add?

If you add a controller, then you can no longer use the `set_commands`; it gets overriden.

> My general point of view is that simple things should be straighforward: moving a robot in position control should be the most straightforward thing to do.

If you put your robot in `servo` or `velocity` mode, then you only need a P value and no real tuning (especially for `velocity` the same small P value works for almost any robot -- of course with velocity, you have infinite torque; servos are similar and I never had issue with tuning them, using a single value for almost all robots).

So basically we can create an example with position control, explaining all the steps: (1) set robot in `servo` or `velocity` mode, (2) add the `PDControl`, (3) use only `P` values, etc... What do you think?

jbmouret:
What I meant was that if I instantiate a controller:
Controller full_pid_controller;
robot.add_controller(&full_pid_controller);

and do my commands: full_pid_controller.set_position()...

Basically, this would be a predefined low-level controller that I can use for higher-level control.

About Velocity: I thought people (you? Pierre?) told me that Talos was not working in velocity mode and that we had to define our PD+ controllers?

costashatz:
> What I meant was that if I instantiate a controller:
> Controller full_pid_controller;
> robot.add_controller(&full_pid_controller);
> 
> and do my commands: full_pid_controller.set_position()...
> 
> Basically, this would be a predefined low-level controller that I can use for higher-level control.

Yes.. This is how the control API works. OK. We should make an issue to create this example.

> About Velocity: I thought people (you? Pierre?) told me that Talos was not working in velocity mode and that we had to define our PD+ controllers?

I never said this. The options are the following:

- Velocity actuators means that no physical constraint is respected (no joint limits, no torque limits, no velocity limits), and the desired velocity is followed always (thus assuming infinite torque available to the motors)
- Servo actuators are the same as velocity actuators but respecting all the physical constraints of the robot. The downside is that each servo is an additional constraint in the optimization loop of DART and thus making thing slower (if you do not have a very powerful machine).
- Torque actuators are the most realistic ones as they are just simulating the forces in joint space. They are also the fastest ones since no additional computation is needed by DART; just forward dynamics (which is already computed). Of course one needs to properly tune the PID values of the low-level controller.

jbmouret:
One of our goals is to provide a good simulators for our robots (Talos and others) that works 'out of the box': we do not want each user to fine-tune his/her talos simulation. We want a Talos/Franka/etc simulator. Like, I want to simulate the Talos but I do not want to think about force vs velocity vs servo. I just want 'something good and validated by the community'.


Then we can provide different options in the Talos/Franka/etc. class ("modes") but we should make it easy...

jbmouret:
By the way, this is one reason to not like this control API ;)

To continue the conversation, I do not think our goal is to provide a library of examples.  20 lines examples are not simple: people (students, engineers, etc.) end up copy-pasting code that they don't understand "just because it was there". Ultimately, people code like it's magic, because they do not have time (and this is alright) to understand all the internal details of a library.

So... if I understand well, if I want to have the different position control modes in the Talos class, then the Talos class cannot use the control API (but this control API would be used internally to do the PIDs). This is fine with me, but this will prevent you from using this class since you like this control API (for reasons that I might understand one day...).

Any solution? I do not think examples for simple position-control is the way to go.

jbmouret:
Question: what is 
```c++
 // First 6-DOFs should always be FORCE if robot is floating base
    for (size_t i = 0; i < 6; i++)
        global_robot->set_actuator_type(i, dart::dynamics::Joint::FORCE);
```
(in the examples)

costashatz:
> One of our goals is to provide a good simulators for our robots (Talos and others) that works 'out of the box': we do not want each user to fine-tune his/her talos simulation. We want a Talos/Franka/etc simulator. Like, I want to simulate the Talos but I do not want to think about force vs velocity vs servo. I just want 'something good and validated by the community'.

There is absolutely no way to get realistic physically-based and accurate simulation without having to tune something. The question is who to tune the simulation. In my mind, I let the user to tune it to their cases. In your mind, you want the community to tune it and provide it. I would not disagree with this option, but you need the community to tune the parameters. I suggest to use torque actuation then, and use the community to tune the PID values. If we have the tuned PID values, then the question is how to provide the tuned parameters and this is easy to do. We can also go with servo actuators as default where it is super simple to have stable P-values (we can then provide them in many different ways).

> people (students, engineers, etc.) end up copy-pasting code that they don't understand "just because it was there". Ultimately, people code like it's magic, because they do not have time (and this is alright) to understand all the internal details of a library.

I tend to agree and disagree at the same time. I know that most of the students code like it's magic, but I do not agree that you need to understand all of the details of a library to not code like it's magic.

In the specific example, they will not even need to know the internal details of the library. Just how to code a PID.

> if I want to have the different position control modes in the Talos class, then the Talos class cannot use the control API (but this control API would be used internally to do the PIDs).

No. You can still use the control API (it depends on the exact implementation). It's the other way around. If you use the control API, then it is not possible to use directly `set_commands()`. Because if you want high-level commands, the robot is not accepting this type of commands. You can of course do: `controller->set_commands()` or `controller->set_parameters()` (as it is now in the naming conventions).

> This is fine with me, but this will prevent you from using this class since you like this control API (for reasons that I might understand one day...).

As I explained above, this is not the case. Also, I am getting a bit mixed signals from your side. From one point of view you do not want automated control API and you want to have `set_commands()`. On the other hand, you want to have a class `Talos` that works out-of-the-box with **any** control mode. If you go for the second option, then `set_commands()` is useless (and I am not talking about the specific implementation but generically since you would like to use the high-level control commands).

> Any solution? I do not think examples for simple position-control is the way to go.

What I was saying is (maybe I did not phrase it correctly):
- For someone that wants position control, the already implemented `PDControl` (for those that want to use the control API) or a simple macro/function that computes a P-controller is sufficient when using the `servo` actuators and `set_commands()`. We can make this crystal clear in the documentation and I said it'd be nice to have a example also. The tuning of this P-value is trivial and we can provide nice default values per robot that we support. Here's what it would look like (for the control API, the `set_commands()` version would be almost identical):

```cpp
talos->set_control_mode("servo");
controller = PDControl(talos->get_servo_p_values());
talos->add_controller(controller);

while(true) {
    controller->set_commands(my_position_commands);
    simu.step_once();
}
```

If we really create separate classes for each robot that we support, then it could even be something like this:

```cpp
talos->set_control_mode("position");

while(true) {
    talos->set_position_commands(my_positions_commands);
    simu.step_once();
}
```

(Similar codes can be shown for the rest)

- For someone that wants to use pure torque control, they can use `SimpleControl` (control API) or just `set_commands()` with `torque` actuators. They can use any high-level controller (e.g., whole body inverse dynamics) in either case.
- For someone that wants position control without the computational overhead of `servo` actuators, we need to provide tuned PD values and then they can use the `PDControl` (control API) or again a macro/function that computes a PD-controller and `set_commands()`. If you have someone to tune the PID values, we can provide nice values for each robot that we support again.
- There is a fourth option to use velocity actuators and P-controllers, but this is identical to the `servo`-position control and will provide less realistic simulations. Could be good for someone doing only animation as it is much faster than using servo actuators.

To be honest, I am not sure how much easier we can make it than this. I do not know if this is what you had in mind, but this is what I meant.

> Question: what is
> 
> ```c++
>  // First 6-DOFs should always be FORCE if robot is floating base
>     for (size_t i = 0; i < 6; i++)
>         global_robot->set_actuator_type(i, dart::dynamics::Joint::FORCE);
> ```
> 
> (in the examples)

This is left-overs from old code. I am literally writing right now a new branch for cleaning up those. But for the record, when you have a floating-base robot if you say that all degrees of freedom (including the base) have servo or velocity actuator type, then the robot stops being floating-base in the sense that you can directly set the velocity of the root joint (first 6 DoFs). In the previous version of robot_dart, when someone was doing `robot->set_actuator_types(type)`, all dofs were overwritten. In the current version by default we do not overwrite mimic joints and the first 6 DoFs if the robot is floating-base (the user has to specify that he wants to overwrite those). But even this should be avoided with what I did in this PR. The user should now do: `robot->set_control_mode("servo");`.

jbmouret:
> There is absolutely no way to get realistic physically-based and accurate simulation without having to tune something. The question is who to tune the simulation. In my mind, I let the user to tune it to their cases.
> In your mind, you want the community to tune it and provide it. I would not disagree with this option, but you need the community to tune the parameters.

 I think the added value of something like robot_dart is in having simulations that are already well tuned so that we do not have to reinvent the simulator for each new project. The fully-flexible version is what Dart does: you can do whatever you want but you have to tune/choose many things.

> I suggest to use torque actuation then, and use the community to tune the PID values. If we have the tuned PID values, then the question is how to provide the tuned parameters and this is easy to do. We can also go with servo actuators as default where it is super simple to have stable P-values (we can then provide them in many different ways).

I am open to all the options for 'realistic'/stable/fast position control...

> > people (students, engineers, etc.) end up copy-pasting code that they don't understand "just because it was there". Ultimately, people code like it's magic, because they do not have time (and this is alright) to understand all the internal details of a library.
> 
> I tend to agree and disagree at the same time. I know that most of the students code like it's magic, but I do not agree that you need to understand all of the details of a library to not code like it's magic.

What I mean is when the example is 20 lines long to do something simple, you simply copy-paste the example and you have no idea what each line does.

> As I explained above, this is not the case. Also, I am getting a bit mixed signals from your side. From one point of view you do not want automated control API and you want to have `set_commands()`. On the other hand, you want to have a class `Talos` that works out-of-the-box with **any** control mode. If you go for the second option, then `set_commands()` is useless (and I am not talking about the specific implementation but generically since you would like to use the high-level control commands).

I am not sure to understand what you mean. We would like to be able to do:
```c++
Talos talos("torque");
//loop
talos.set_commands(...); //torque commands

Talos talos2("position");
//loop
talos2.set_commands(...); // positions

> * For someone that wants position control, the already implemented `PDControl` (for those that want to use the control API) or a simple macro/function that computes a P-controller is sufficient when using the `servo` actuators and `set_commands()`. We can make this crystal clear in the documentation and I said it'd be nice to have a example also. The tuning of this P-value is trivial and we can provide nice default values per robot that we support. Here's what it would look like (for the control API, the `set_commands()` version would be almost identical):
> 
> ```c++
> talos->set_control_mode("servo");
> controller = PDControl(talos->get_servo_p_values());
> talos->add_controller(controller);
> 
> while(true) {
>     controller->set_commands(my_position_commands);
>     simu.step_once();
> }
> ```

This is not bad, but I would like to hid the PDController.

> If we really create separate classes for each robot that we support, then it could even be something like this:
> 
> ```c++
> talos->set_control_mode("position");
> 
> while(true) {
>     talos->set_position_commands(my_positions_commands);
>     simu.step_once();
> }
> ```

Yes, this is what I mean. But that means having "set_position_commands" and "set_commands". Which is not very elegant... Or we do not use 'modes' and use set_torque_command, set_velocity_command, etc.

> * For someone that wants position control without the computational overhead of `servo` actuators, we need to provide tuned PD values and then they can use the `PDControl` (control API) or again a macro/function that computes a PD-controller and `set_commands()`. If you have someone to tune the PID values, we can provide nice values for each robot that we support again.

How big is this overhead? I have nothing against using proper servos if this is not unstable and not too slow... I was simply suggesting these PD controllers because this is what is currently used iCub & Talos. This is not necessarily the right approach (I did not do this...).

**Comment by costashatz** [comment]
> Yes, this is what I mean. But that means having "set_position_commands" and "set_commands". Which is not very elegant... 

If we have a separate class for each robot that we support, it is possible to just do `set_commands` and not `set_position_commands`. Although to be honest I like when control commands are explicit, otherwise people tend to forget what values they send.

> Or we do not use 'modes' and use set_torque_command, set_velocity_command, etc.

This is going to be complicated to actually implement. Because it would mean that you can change the control mode of each dof on the fly. I am not sure if we really want that.

> How big is this overhead?

Loosely speaking, the running time of their solver is O(n^3) where n is the number of constraints (contacts, servo, joint limits, etc.). The more constraints you have, the more time it takes. Gladly, `servo` constraints are not as costly as contact constraints since they are easily satisfied, but still they do make the solver slower. I haven't really timed it, but we can.

DART also supports iterative solvers (that should be faster), but last time that I used them they were not working very well. It's been a long time though.

In any case, this PR is just creating an alias for `set_actuator_type` as to be DART-free. We should move the discussion on whether we should create separate classes for each robot that we support or not (if we do not have separate classes per robot, then the current state to my opinion is the easiest option without losing generality) to another issue.

**Comment by jbmouret** [comment]
So... what do we do? I think we do not have a clear consensus here.
- either we do not set PD+force control as "a control mode" (which is OK if this is something we use very rarely and not the "default" way to control a robot in position) --- in that case we instantiate a special controller and do our stuffs
- or we have an additional mode in some robot classes (e.g. Talos) that instantiate the PD/force with the right parameters, which is called using set_position_commands or something like this, but why not having this in Robot.cpp in that case.

I guess the true question is wether it is actually better to use the servo actuation for our robots (given that we want to have fast simulations). It is good that the servo joints do not have parameters to tune, of course.

**Comment by jbmouret** [comment]
I tried a small "benchmark":
- Talos
- 2 joints are controlled (at 100Hz)
- 1000Hz for Dart
- 10 seconds simulation
- collision: FCL

```
robot->set_control_mode("velocity");
```
=> 9.5 seconds (without a PID, but this should not change anything on the running time)

```
robot->set_control_mode("servo");
```
=> 20 seconds

**Conclusion**: 
- velocity mode = real-time
- servo mode = 2 times slower than real-time!

So, yes, there is a large overhead.

costashatz:
> So... what do we do?

I thought about it a bit and here are my thoughts:

- "Control_modes" is not the best choice of words for what I am doing in this PR. I will change this PR to have better names.
- I do not believe that providing a `control_mode` to the generic robot is easy to do. This is because we need to be generic (`Robot` is a very generic class) and it'll be difficult to catch all cases.
- We can provide classes for specific robots and there we can have some kind of `control modes`. Because they will depend on the specific robots: e.g. a walking controller for a hexapod.

> * collision: FCL

I would avoid FCL. Go either for Bullet (they have a small memory leak, but maybe they fixed it) or if you have simple shapes, use DART's simple collision detector (I always convert everything to simple shapes for collision and use this one as it is the most deterministic one and the fastest one). FCL is super slow as it generates 5-6 contacts per pair of objects (sometimes even more!!!) and this makes it even slower for the simulation.

> * velocity mode = real-time

Velocity mode is ignoring all the physical limits. They are basically only kinematic joints: no joint limits, no torque limits, no other constraints. If you want to simulate realistic robotic setups, you should avoid velocity joints. They are intended for animation or simple verifications.

> * servo mode = 2 times slower than real-time!

This is normal for big robots like Talos; imagine that each servo motor is imposing 6-8 constraints to the solver. We can also trying to switch to an iterative solver, that is supposed to be faster, instead of the LCP one (that is the default one).

> * torque mode = 2 times slower than real-time (in that specific case)

This is weird. I guess this comes from the collisions. The speed should be almost the same with velocity actuators. I've benchmarked those quite a few times and they are more or less equivalent. So I guess this is coming from the collisions.

> * => for maximum simulation speed, we should use velocity mode (which would be like a soft constraint in Dart). I need to test more for the torque mode, but it does not seem obviously faster.

As I said, velocity actuators are only for simple verification or kinematic animations. They are not intended for accurate physics simulation. Also torque-mode should be as fast as velocity actuators; try with a PID and you should be getting very similar run-times.

So overall my conclusion is that we should have torque mode with well-tuned PID controllers, and velocity mode for simple things and verification. In velocity mode it is also impossible to do inverse dynamics control (except if you convert the computed torque to positions or velocities, which is not that good to do), whereas in torque-mode you give this possibility.

jbmouret:
I am fine to put this in robot-specific classes. However, don't we need similar things everywhere (besides de parameters)?

costashatz:
> However, don't we need similar things everywhere (besides de parameters)?

Maybe. Let's try to see what exactly we need and how it needs to be implemented. If it can be generic, I am fine with putting it inside the `Robot` class. Although as a principle I would like to keep separated the `control` part from the `robot` part, but I can "put some water in my wine" if this can be very generic and it'll help people using the library.

jbmouret:
With the Bullet collision this is much much slower once the robot hits the ground (like, it did not finish after 1 minute...).

jbmouret:
> > However, don't we need similar things everywhere (besides de parameters)?
> 
> Maybe. Let's try to see what exactly we need and how it needs to be implemented. If it can be generic, I am fine with putting it inside the `Robot` class. Although as a principle I would like to keep separated the `control` part from the `robot` part, but I can "put some water in my wine" if this can be very generic and it'll help people using the library.

Do torque joints comply with joint limits/velocities/etc. ? If not, we need to add this in addition to the PD / SPD part.

costashatz:
> Do torque joints comply with joint limits/velocities/etc. ?

Yes they do. This is the default and most realistic way of simulating robots in DART.

> With the Bullet collision this is much much slower once the robot hits the ground (like, it did not finish after 1 minute...).

Good to know that they did not fix it yet.

costashatz:
@jbmouret I think this is ready to be merged.. Let me know if you have any issue/comment (we should handle the robot-specific control modes, etc. in a different PR).

jbmouret:
Good for me!

jbmouret:
@dalinel @PedroDesRobots Can you give me the iCub urdf/meshes that we use with Dart?

PedroDesRobots:
@jbmouret you can find here : 

**icub_description** package : https://gitlab.inria.fr/larsen-robots/icub_description
urdf : https://gitlab.inria.fr/larsen-robots/icub_description/-/blob/master/urdf/icub.urdf

jbmouret:
@PedroDesRobots and the xacro that corresponds to it? what do we need to take?

jbmouret:
(btw. sorry, I thought this was the other pull request)

PR #83, opened 2020-06-07
Possibility of displaying 3D axes

costashatz:
This PR makes a few minor fixes in the `GuiData` and adds the possibility of display 3D axes. Here's an example:

image

These changes require to use the `Magnum EigenIntegration` but this is harmless as we just need to do just one `-DWITH_EIGEN=ON` when building `magnum-integration`. With a small restructure, now it is possible to more easily add more debug drawings if needed in the future.

### DIFF
--- a/ci/install_magnum.sh
+++ b/ci/install_magnum.sh
@@ -34,7 +34,7 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
     git clone https://github.com/mosra/magnum-integration.git
     cd magnum-integration
     mkdir build && cd build
-    cmake -DCMAKE_BUILD_TYPE=Release -DWITH_DART=ON ..
+    cmake -DCMAKE_BUILD_TYPE=Release -DWITH_DART=ON -DWITH_EIGEN=ON ..
     make -j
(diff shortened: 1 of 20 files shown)


jbmouret:
I like it! Is there a way to easily show what is x/y/z ? we could draw the letters like the arrows (these are easy letters)

costashatz:
> So, I guess we is not straightforward to add

It is not too difficult to add, but I will leave this along with the PR for adding text rendering.

> we should at least document the color coding.

Yes, XYZ = RGB. We will put this in the docs.

One question is do we want the axes to have proper depth or should they be rendered always on top of everything else? Right now they are going through the depth checking, but I guess since this is for debugging purposes having them on top always might be a better idea? What do you think?

jbmouret:
I think this is fine like it is. This might look weird if they do not go through the depth checking.

costashatz:
**Merging...**

costashatz:
@jbmouret if we put something like this (we could change the colors of each axis also)? Do you prefer it over lines and text?

image

jbmouret:
This is good for me.

PR #90, opened 2020-08-17
Αdd easy limits access and reset in Robot class

costashatz:
Αdd easy limits access and reset in Robot class (+ python fix)..


costashatz:
**Merging...**

PR #91, opened 2020-08-18
Add jacobian, mass matrix and other helpers in Robot

costashatz:
Add jacobian, mass matrix and other helpers in Robot. A few things to note:

- I have given the possibility of "cutting" a jacobian or a mass matrix to a view of only the DOFs of interest.
- Although this might be good for debugging or very specific purposes, you should not use these views to perform computations for control/estimation (if you do not know exactly what you are doing, you will most probably do everything wrong).
- You should do something like the following: `joint_torques = robot->vec_dof(robot->jacobian("my_end_effector").transpose()*end_effector_wrench);`.


costashatz:
**Merging...**

PR #92, opened 2020-09-01
Fast Talos

jbmouret:
- new URDF "talos_fast"
- new example

This is a talos with simplified collision detection to make things faster.


costashatz:
We might want to add a version of Talos without the upper body (maybe someone wants to focus only on the locomotion of a biped):
taloss

What do you think?

jbmouret:
If you use it, no problem for me. The main challenge will be to keep synchronized all these URDFs and documenting them (what they do, how they differ, etc.).

jbmouret:
Should we merge or should we add your lower body?

jbmouret:
Also, what do you think of starting to add Readme.md in the directories of the robot? like, what are the features/trade-offs  of each URDF.

costashatz:
> Should we merge or should we add your lower body?

I think we should not add this lower body as I agree with you that it'd be difficult to maintain all the different models.

> Also, what do you think of starting to add Readme.md in the directories of the robot? like, what are the features/trade-offs of each URDF.

Yes. This is a good idea.

costashatz:
> Also, what do you think of starting to add Readme.md in the directories of the robot? like, what are the features/trade-offs of each URDF.

Should we do this in this PR? Or should we create a new one?

jbmouret:
I should quickly do it in this PR.

Btw, LAAS has a "fast talos" URDF too (with boxes or automatically generated boxes). We should check it.

costashatz:
**Merging**

PR #93, opened 2020-09-07
Make boost.process optional (but make it possible to record videos)

jbmouret:
I implemented it in the "good old way" because the unbuntu that is on the official Talos dev dockers/PC (16.x) does not have the right boost version.

We might remove this in the future once everybody has a a recent boost.

### DIFF
--- a/src/examples/cameras.cpp
+++ b/src/examples/cameras.cpp
@@ -26,8 +26,8 @@ int main()
 {
     std::srand(std::time(NULL));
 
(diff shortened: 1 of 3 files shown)


costashatz:
> We might remove this in the future once everybody has a a recent boost.

Why do we need boost.process if we can do it without it? Is there any real need for boost process?

jbmouret:
The boost.process code is much more elegant, portable and I think robust.  We already have a dependency to boost, so this is essentially "for free" (except on old Ubuntu).

PR #94, opened 2020-09-07
improve example of icub

jbmouret:
Play a bit with the iCub example so that it moves (a kind of squat)


costashatz:
This should be ready to be merged. Once CI tests pass, I'll merge..

PR #96, opened 2020-09-17
Robot reset

costashatz:
This PR makes the `robot.reset()` function better and makes sure that when we clone or initialize with a skeleton, we reset everything to get a clean robot structure.


costashatz:
**Merging...**

Issue #79, opened 2020-06-05
Create a Sensor API


costashatz:
As discussed in #71 we should create a `Sensor` API.

PR #97, opened 2020-09-19
API for generic Sensors

costashatz:
This PR intends to add a generic API for creating and using sensors in robot_dart. This will also remove `CameraOSR` class, and adds a Magnum `Camera` class as a sensor. Things remaining:

- ~~Remove `CameraOSR`~~
- ~~Add python bindings~~
- ~~Add tests for sensors (only camera now)~~ **Tests for IMU and FT sensors added** (no test for camera can be made at the moment)
- ~~Add a simple IMU sensor~~
- ~~Add force/torque sensor~~
- ~~Body Node name does not suffice. We need to have reference to the robot.~~


src/robot_dart/gui/magnum/base_application.hpp
```diff
@@ -131,7 +131,9 @@ namespace robot_dart {
                 void update_lights(const gs::Camera& camera);
                 void update_graphics();
                 void render_shadows();
+                // This should not be used
```
jbmouret, line 134: why?

costashatz: Because the function with just the name would stop on the first `BodyNode` or `ShapeNode` with the given name. This is OK if you have only one robot in the world or all the body nodes/shape nodes have different names. This is not always the case, since body nodes and shape nodes are allowed to have the same name if they are under different skeletons (robots). I meant to delete this function as I created one with a pointer to the body node; I forgot. I will delete it.

src/python/example.py
```diff
@@ -60,12 +60,12 @@ def __call__(self):
...
-# Add a camera to the end-effector of the manipulator
-camera = rd.gui.CameraOSR(simu, graphics.magnum_app(), 256, 256)
+# Add a camera sensor to the end-effector of the manipulator
+camera = rd.sensor.Camera(simu, graphics.magnum_app(), 256, 256)
 rot = dartpy.math.AngleAxis(3.14, [1., 0., 0.]).to_rotation_matrix()
 rot = rot.dot(dartpy.math.AngleAxis(1.57, [0., 0., 1.]).to_rotation_matrix())
-camera.attach_to("arm_link_5", dartpy.math.Isometry3(rotation=rot, translation=[0., 0., 0.1]))
-simu.add_camera(camera)
+camera.attach_to_body(robot.body_node("arm_link_5"), dartpy.math.Isometry3(rotation=rot, translation=[0., 0., 0.1]))
+simu.add_sensor(camera)
```
jbmouret, line 68: What happens if we add a sensor without attaching it to a body first?

src/examples/cameras.cpp
```diff
@@ -54,7 +54,7 @@ int main()
     graphics->record_video("video-main.mp4", simu.graphics_freq());
 
     // Add camera
-    auto camera = std::make_shared<robot_dart::gui::magnum::CameraOSR>(&simu, graphics->magnum_app(), 256, 256);
+    auto camera = std::make_shared<robot_dart::gui::magnum::sensor::Camera>(&simu, graphics->magnum_app(), 256, 256);
```
jbmouret, line 57: It sounds weird to have a sensor in the gui namespace...

src/python/example.py
costashatz: It works as expected.. You can even attach it to a different body node while the simulation is running..

src/examples/cameras.cpp
costashatz: This is true.. I'll try to make the camera sensors headers to be independent of `Magnum` and only the `cpp` files to require `Magnum`.

costashatz:
@jbmouret the first version of the sensor API should be completed. This PR fixes #79. I'll merge once the CI tests pass.

costashatz:
**Merging..** I note that we should create a random generator (similar to limbo) and add some noise models to the sensors...

PR #100, opened 2020-09-26
Create a robot pool

### LINKED ISSUE #99: Include a clone pool
We use this pattern often (to avoid creating a new robot for each evaluation). We could have a polished implementation in robot_dart.

jbmouret:
This is useful when we use robot_dart in many threads and we want to avoid parsing new robots from scratch.

Fix #99

### DIFF
--- a/src/examples/robot_pool.cpp
+++ b/src/examples/robot_pool.cpp
@@ -1,0 +1,76 @@
+#include <algorithm>
+#include <cstdlib>
+#include <iostream>
+#include <thread>
+
+#include <robot_dart/robot_dart_simu.hpp>
+#include <robot_dart/robot_pool.hpp>
+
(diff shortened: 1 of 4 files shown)


costashatz:
@jbmouret thanks for this. I pushed a fix for a missing linking flag, but I have one bigger concern. What happens if we want to use to use two robots (and thus there is a need for 2 robot pools with different robots)? The way it is now, this is not possible..

jbmouret:
Good question: 
- we can choose to not cover this case
- we can suppress the singleton access (it is handy but the user can do a global variable instead or link instances)

I would vote for the second case.

costashatz:
> I would vote for the second case.

Agree!

costashatz:
**Merging...**

Issue #41, opened 2020-05-10
Add API for drawing helper text/2D visualizations


costashatz:
We should add helper functions for the user to easily add text and 2D visualizations on the screen. Here's a list:

- Text that can be changed at every time-step
- 2D rectangles with colors/fillings
- ...

@jbmouret @dalinel @PedroDesRobots let me know if you have something to add..

costashatz:
@jbmouret @dalinel @PedroDesRobots would it make sense to have a more complete UI support (e.g., like [this](https://magnum.graphics/showcase/ui-gallery/))? I think that we would need mainly only text, so it'll be too much to incorporate properly a UI (Magnum has a built-in, but we would have to expose the generic API somehow).. The text-only is quite easy to do.. Let me know what you think..

jbmouret:
I think some text is enough but if we get something nicer for free, why not.

costashatz:
> I think some text is enough

I think I will do that so that we can have something quickly.

> if we get something nicer for free, why not.

The built-in Magnum UI does not come for free, it just comes easier than ImGui (and less dependencies). Why would we need a full UI though? Do we want to expose some user interaction? Although we could to be honest :P

jbmouret:
If there are additional dependencies, I think we should stay with the text. If this is already a component of Magnum that we simply have to activate, then we could use it as it would allow more in the future (e.g., press a button to have a random force that pushes the robot).

jbmouret:
Any news of this? it would be nice to have a minimal 'dashboard' about the simulation (at least the time like Gazebo). Most things are robot-specific but we could provide a Talos-dashboard, for example, with the values of the sensors (IMU, force sensors of the feet, for instance).

costashatz:
> Any news of this? it would be nice to have a minimal 'dashboard' about the simulation (at least the time like Gazebo).

No news. The easiest way would be to use the (somewhat limited, but easy to use) `Magnum-UI` which is another library but can be seen as a Magnum component (and thus not really a different dependency). Let me see if I can do this before the end of July and get back to you..

> Most things are robot-specific but we could provide a Talos-dashboard, for example, with the values of the sensors (IMU, force sensors of the feet, for instance).

These specific things should go to your Talos code. We should however provide a relatively simple API to add those things. But for this, I should probably first finish the Sensor API (see #79).

costashatz:
This is partly fixed by #98, but leaving this open for adding the rest of the visualizations.

jbmouret:
[...]

PR #101, opened 2020-09-28
Define placeholders for text

jbmouret:
See #41 
- status bar
- text panel (top left)

Still a bug with the vertical position of the status bar when we resize the window when it is too wide.

### DIFF
--- a/src/examples/talos.cpp
+++ b/src/examples/talos.cpp
@@ -59,8 +59,7 @@ int main()
             Eigen::VectorXd commands = (init_positions + delta_pos) - robot->positions(dofs);
             robot->set_commands(commands, dofs);
         }
-        simu.set_text_panel(robot->model_filename() + "\n"
-            + simu.default_text_panel());
+
         simu.step_world();
     }
 
--- a/src/examples/talos.cpp
+++ b/src/examples/talos.cpp
@@ -59,7 +59,8 @@ int main()
             Eigen::VectorXd commands = (init_positions + delta_pos) - robot->positions(dofs);
             robot->set_commands(commands, dofs);
         }
-        simu.set_text_panel("TALOS simulation\n" + simu.default_text_panel());
+        simu.set_text_panel(robot->model_filename() + "\n"
+            + simu.default_text_panel());
         simu.step_world();
     }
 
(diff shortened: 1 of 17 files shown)


src/robot_dart/robot_dart_simu.cpp
```diff
@@ -407,18 +414,48 @@ namespace robot_dart {
...
+
+    void RobotDARTSimu::set_status_bar(const std::string& str) {
+        if (_status_bar)
+            _status_bar->text = str;
+    }
+    
+    std::string RobotDARTSimu::default_text_panel() const {
+        std::ostringstream out;
+        out.precision(3);
+        double rt = _scheduler.current_time();
```
costashatz, line 445: `_scheduler.current_time()` does not give the real time passed. At the moment, only `_scheduler.step()` returns this.

src/examples/talos.cpp
```diff
@@ -59,6 +59,8 @@ int main()
             Eigen::VectorXd commands = (init_positions + delta_pos) - robot->positions(dofs);
             robot->set_commands(commands, dofs);
         }
+        simu.set_text_panel(robot->model_filename() + "\n"
+            + simu.default_text_panel());
```
costashatz, line 63: It is a bit weird to have to re-set the text just for the simulation time to be updating. We could add a boolean (whether to display the default text or not), and it should be added automatically. What do you think?

src/robot_dart/robot_dart_simu.cpp
jbmouret: Should we update the scheduler, so?

src/examples/talos.cpp
jbmouret: Why not. I kinda like the idea to "make your own" : here you can for instance add a title above the default (if you do not call set_text_panel, then you have the default). But I am fine with a Boolean if you prefer.

src/robot_dart/robot_dart_simu.cpp
costashatz: There's no other way to know the real-time factor. We need to call `scheduler.step()`; we could save this value though to be able to access it..

jbmouret: Yes, this is what I had in mind. I will do it.

PR #102, opened 2020-10-02
Robot classes (WIP)

jbmouret:


jbmouret:
@costashatz should we make one class for each robot (iiwa, franka, etc.). They would be empty (? maybe some torque sensors?) but that would be more consistent. What do you think?

costashatz:
> @costashatz should we make one class for each robot (iiwa, franka, etc.). They would be empty (? maybe some torque sensors?) but that would be more consistent. What do you think?

I think that we should add classes only to robots that require some "special manipulation". E.g., franka has the gripper and also needs a special setting for the material in order to be drawn nicely.

On the other hand, this would not be consistent. Let me think about this a bit, and come back at you.

costashatz:
@jbmouret only python bindings are missing.. I will do them tomorrow and merge..

PR #106, opened 2020-10-06
fix the real time factor + improve scheduler + improve status bar

jbmouret:
We were computing the real-time factor using the wall time instead of the iteration time. This means that if we were real-time but had one iteration slower than realtime, we could never catch up and reach a real-time factor of 1. Basically, to reach a real-time factor of 1, we needed many iterations faster than real time if we had been late in the past.

I switched to a real-time factor that corresponds to the iteration time, which corresponds better to what we want to know (I think).

By the way, if you log the iteration time, you will see that in graphics we often have iterations that last 14 ms (400 ms at the start), which do not come from the 'sync' mechanism (apparently). There is no problem in _plain.

### DIFF
--- a/src/examples/iiwa.cpp
+++ b/src/examples/iiwa.cpp
@@ -39,6 +39,7 @@ int main()
     simu.add_checkerboard_floor();
     simu.add_robot(global_robot);
     simu.add_robot(ghost);
+    simu.set_text_panel("IIWA simulation");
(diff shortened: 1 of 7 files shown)


costashatz:
@jbmouret is this ok to merge?

jbmouret:
Yes. I will do a different PR for the next changes.

PR #109, opened 2020-10-07
Torque sensor implementation + iiwa example

ivanbergonzani:


src/robot_dart/sensor/torque.cpp
```diff
@@ -0,0 +1,134 @@
...
+            Eigen::Vector6d F2 = Eigen::Vector6d::Zero();
+            auto child_body = _joint_attached->getChildBodyNode();
+            // ROBOT_DART_ASSERT(child_body != nullptr, "Child BodyNode is nullptr", );
+            if (child_body)
+                F2 = -dart::math::dAdT(_joint_attached->getTransformFromChildBodyNode(), child_body->getBodyForce());
+
+            // We always compute things in SENSOR frame (aka joint frame)
+            _wrench = (_direction == "parent_to_child") ? F2 : -F2;
+
+            Eigen::Matrix3d R = _joint_attached->getTransformFromChildBodyNode().linear().transpose();
```
costashatz, line 55: Why do we need this? F2 is already in joint space. We need it back in `ChildBodyNode` coordinates? Why?

ivanbergonzani: the implementation was based on the FT sensor class.
yes line 53 and below should be useless for the torque sensor

costashatz:
@ivanbergonzani @jbmouret after our private discussions, I think this is ok to be merged, no?

PR #110, opened 2020-10-08
Franka sim

dinies:
### Franka simulation example

**added** : 

- urdf model of Franka robot
- meshes related to the model
- example file franka.cpp to simulate the robot

**notes**
~~- for the time being, the robot stays still.~~
~~- there an additional blank line on the iiwa.cpp that I forgot to revert before pushing the changes~~


wscript
```diff
@@ -139,7 +139,7 @@ def configure(conf):
     
     # add the prefix
     conf.env['CXXFLAGS'] = conf.env['CXXFLAGS']
-    print(conf.env['CXXFLAGS'])
+    print(conf.env)
```
costashatz, line 142: This is huge to print. Can you revert this?

costashatz:
@dinies thanks for this PR.

@jbmouret @dinies I updated the URDF to one with the hand. Also this one that I had has completely different masses than the one you had. We should use one URDF that we know it makes sense.

Do you want to create 2 URDFs? One with the hand and one without?

Note: Ignore the weird shading of the first 2 links. I am creating a different PR to handle this.

jbmouret:
We need the gripper, so no need for a second URDF.

We can merge for now if you like. I still want to ask around for "the best URDF" according to our colleagues, but we can change the URDF later.

costashatz:
**Merging...**

PR #112, opened 2020-10-10
Fix specularity

costashatz:
We were having too much specularity in our scenes. This PR addresses this issue + makes a small fix for ill-formed materials. Now there is a parameter to tune how much specularity we want in our scene.


jbmouret:
Thanks! It looks much nicer. OK for merging asap.

PR #115, opened 2020-10-28
ForceTorque sensor stores wrenches/IMU sensor improvement

costashatz:
ForceTorque sensor stores and gives wrenches (in addition to force/torque vectors). One note: force/torque getters make copies now..


costashatz:
IMU sensor returns the angular position. For now this is the ground truth value; in subsequent versions we need to do this closer to what real sensors do.

jbmouret:
OK for merging asap.

costashatz:
> OK for merging asap.

I'll wait for the CIs to pass..

Issue #116, opened 2020-10-28
Contact friction (i.e. mu)


limbryan:
Hi, 

I was wondering if there were any existing simple functions or ways to define the friction between the robot and the floor (or for contact friction in general). This is pretty important for legged robots as the friction can heavily affect learnt behavior especially. 

I have looked through robot_dart and only found set_cfriction_coeffs and some similar functions but there was little documentation and examples surrounding it so it was unclear what it was actually for - also noticed it used the DoF to implement the friction so likely that it could have been joint friction. I ended up implementing this myself using dart functions (setFrictionCoeff) like below: 

![image](https://user-images.githubusercontent.com/46229436/97452664-e23f8080-192c-11eb-97eb-e3a799d7ba15.png)

but does not seem very elegant to be selecting all the nodes of the floor and setting the coeff of friction. Would it be worth implementing a simple function that comes with robot_dart which just takes in a coeff of friction between 0 and 1 and automatically applying them to the contacts between robot and floor?

costashatz:
Hello @limbryan,

Thanks for using our library!

Indeed you are right, we do not have any function for this. Let me fix this as soon as possible (I'll try to do this, this week).

limbryan:
Hi @costashatz , 

Thank you for the quick response and for this great tool. Great, I will check back in next week. Thank you again!

costashatz:
@limbryan I found some time today and here's the PR #117...

One important thing to note is that the default `FrictionDirection` is a zero vector and thus the friction coefficients are ignored! So one needs to set this at least for one BodyNode involved in the collisions! I will attempt to create an automatic way of doing this for the floor, so that at least everything that interacts with the floor gets the default friction coefficients instead of collision without coefficients. We should definitely put this in the docs (speaking to myself, no worries).

Let me know if this works for you..

limbryan:
Thanks @costashatz! That was really fast! I went through the commit changes quickly and noticed that the python wrappers for these changes might have been missed out as well. Might be worth to keep this in mind for consistency. 

I did not know that about the FrictionDirection. This likely means that my implementation did not really work as I did not set the direction - though I did see a change in the behaviors learnt which is weird and the coverage of the map.  What is the definition of friction direction (realistically, we would have the same friction coefficient applied in all directions no?) does this end up meaning the friction cone? and I guess the big question is should this direction be determined by the user for generic purpose, or is this what you meant by this should automatically be taken care of included once it is done for the floor?  

Thanks again btw!

[...]

PR #117, opened 2020-10-28
Friction/restitution parameters available, floor as a robot

costashatz:
This is meant to solve #116... I have exposed almost all contact parameters for each BodyNode, and made the floor as a robot so that we can alter its values using the same API.


costashatz:
@limbryan let me know if this is good for you, so that I can merge.. Thanks!

limbryan:
@costashatz this works great! I have tested it out as well and the api is pretty straightforward so it looks good to me. I did not get to test out the some of the other functions as I am on an older version of dart but the other ones worked well for me. I think it is good to merge! Thanks again - this was done very quickly!

PR #119, opened 2020-11-03
Make it possible to change the font size of the panel & status bar

jbmouret:


jbmouret:
(the python side is probably not working for now)

src/robot_dart/gui/magnum/base_application.hpp
```diff
@@ -107,6 +107,9 @@ namespace robot_dart {
                 // These options are only for the main camera
                 bool draw_main_camera = true;
                 bool draw_debug = true;
+
+                // text
+                int font_size = 32;
```
costashatz, line 112: It seems to me that this is not used anywhere...

src/robot_dart/gui/magnum/base_application.cpp
```diff
@@ -176,7 +176,7 @@ namespace robot_dart {
 
                     /* Glyphs we need to render everything */
                     /* Latin characters for now only */
-                    _glyph_cache.reset(new Magnum::Text::DistanceFieldGlyphCache{Magnum::Vector2i{2048}, Magnum::Vector2i{512}, 22});
+                    _glyph_cache.reset(new Magnum::Text::DistanceFieldGlyphCache{Magnum::Vector2i{2048}, Magnum::Vector2i{512}, 32});
```
costashatz, line 179: Why did you change the size here?

src/robot_dart/robot_dart_simu.hpp
```diff
@@ -118,14 +119,14 @@ namespace robot_dart {
 
         simu::GUIData* gui_data();
 
-        void enable_text_panel(bool enable = true);
+        void enable_text_panel(bool enable = true, float font_size = -1);
```
costashatz, line 122: In the robot_dart API everything is a double. We should do this as a double and cast it if necessary in other parts of the code.

src/robot_dart/gui/magnum/base_application.cpp
jbmouret: I wanted to have better glyphs, but it seems that it did not change anything.

src/robot_dart/robot_dart_simu.hpp
jbmouret: OK. Double is fine.

costashatz:
@jbmouret thanks for this! I did some small final touches. I'll wait the CIs and merge..

PR #122, opened 2020-11-27
Remove the link to the dependencies (esp. assimp)

jbmouret:
It seems that linking to the dependencies is not required with modern OS/linkers (that is, we link with Dart and Dart links with assimp on its own).

If we do it, this easily leads to inconsistencies if we do not find the same version (e.g dart uses assimp5 and we find assimp3). Also, this is useless code to maintain :)

### DIFF
--- a/waf_tools/dart.py
+++ b/waf_tools/dart.py
@@ -40,19 +40,6 @@ def check_dart(conf, *k, **kw):
             includes_check = [os.environ['RESIBOTS_DIR'] + '/include'] + includes_check
             libs_check = [os.environ['RESIBOTS_DIR'] + '/lib'] + libs_check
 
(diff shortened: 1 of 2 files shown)


costashatz:
@jbmouret thanks for this. I will need to adapt the Magnum part. If the plugin is statically compiled, we need to link to assimp and the other libraries. Or not? I think that we do..

jbmouret:
I think we need to test because it might work without adding the dep.

costashatz:
@jbmouret thanks for this. I put back the checks in `magnum_plugins` as they are not affecting the linking flags. As soon as the CIs pass, I will merge. Let me know if there is any issue with my new commit..

PR #123, opened 2020-12-08
Add the possibility to change the background color

jbmouret:
- we also slightly alter (lighter) the color of the checkered floor
- example in talos.cpp

For now, I did not do the change for the sensor/camera.cpp as there is no graphics configuration here (it would require to add an argument to the constructor).


costashatz:
@jbmouret thanks for this! I did some small final touches. I also reverted the change of the checkerboard color and made the two colors parameters so that someone can change them. I'll wait the CIs and merge..

PR #125, opened 2020-12-17
Shadow per Light

costashatz:
Now lights can decide to not cast shadows. One of the default lights now does not cast a shadow (removing thus the double shadows that seem weird).


costashatz:
image

jbmouret:
It compiles and runs on Mac but we previously had the same behavior (single shadow).

costashatz:
> It compiles and runs on Mac but we previously had the same behavior (single shadow).

Yes Mac is abandoning OpenGL. We should switch to Vulkan when Magnum has it ready (soon I guess).

PR #129, opened 2021-03-08
Add full RGB-D sensor capabilities

kostastsing:
~~In this pull request, I have added a function that calculates the intrinsic matrix of a camera from the perspective projection matrix.~~

http://ksimek.github.io/2013/06/03/calibrated_cameras_in_opengl/

This PR adds capabilities to the camera sensor in order to get the point cloud from the depth buffer.

### DIFF
--- a/src/robot_dart/gui/magnum/gs/camera.cpp
+++ b/src/robot_dart/gui/magnum/gs/camera.cpp
@@ -71,10 +71,10 @@ namespace robot_dart {
 
                 Magnum::Matrix3 intrinsicMatrix()
                 {
-                    Magnum::Matrix4 perspProj = Magnum::Matrix4::perspectiveProjection(_fov, _aspect_ratio, _near_plane, _far_plane);
-                    return {{perspProj[0][0], 0., 0.},
-                            {perspProj[0][1], perspProj[1][1], 0.}
-                            {perspProj[0][2], perspProj[1][2], -1.}};
+                    Magnum::Matrix4 projection = _camera->projectionMatrix();
+                    return {{projection[0][0], 0., 0.},
+                            {projection[1][0], projection[1][1], 0.}
+                            {projection[2][0], projection[2][1], -1.}};
                 }
 
                 Camera3D& Camera::camera() const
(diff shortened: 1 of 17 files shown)


src/robot_dart/gui/magnum/gs/camera.cpp
```diff
@@ -69,6 +69,14 @@ namespace robot_dart {
 #endif
                 }
 
+                Magnum::Matrix3 intrinsicMatrix()
+                {
+                    Magnum::Matrix4 perspProj = Magnum::Matrix4::perspectiveProjection(_fov, _aspect_ratio, _near_plane, _far_plane);
```
costashatz, line 74: Why not using the `_camera->projectionMatrix()`? See [here](https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1Camera.html#a21f5b5c5f4aa7726a6343328e065aa43).

src/robot_dart/gui/magnum/gs/camera.cpp
```diff
@@ -69,6 +69,14 @@ namespace robot_dart {
...
+                    return {{perspProj[0][0], 0., 0.},
```
costashatz, line 75: Not sure this initialization is correct. `Matrix3` accepts 3 column vectors (one for each column). Also the `[]` operator first access columns, then rows..

src/robot_dart/gui/magnum/gs/camera.hpp
```diff
@@ -52,6 +52,7 @@ namespace robot_dart {
                     Magnum::Float fov() const { return static_cast<Magnum::Float>(_fov); }
                     Magnum::Int width() const { return _camera->viewport()[0]; }
                     Magnum::Int height() const { return _camera->viewport()[1]; }
+                    Magnum::Matrix3 intrinsicMatrix();
```
costashatz, line 55: This should have the following signature: `Magnum::Matrix3 const intrinsic_matrix()` to be consistent with the style.

src/robot_dart/gui/magnum/gs/camera.hpp
```diff
@@ -52,7 +52,7 @@ namespace robot_dart {
                     Magnum::Float fov() const { return static_cast<Magnum::Float>(_fov); }
                     Magnum::Int width() const { return _camera->viewport()[0]; }
                     Magnum::Int height() const { return _camera->viewport()[1]; }
-                    Magnum::Matrix3 intrinsicMatrix();
+                    Magnum::Matrix3 const intrinsicMatrix();
```
costashatz, line 55: It should still be `intrinsic_matrix()` for consistency.. ;)

PR #132, opened 2021-04-21
Updates installation instructions.

matthias-mayr:
- Default build type of dependencies "release"
This is taken from CI and I argue that it's a good default for most users.
 - Add missing "DWITH_STBTRUETYPEFONT=ON"
This was missing when building magnum plugins and prevented the build of graphical examples.
 - Instruction to build the examples

### DIFF
--- a/ci/install_dart_latest.sh
+++ b/ci/install_dart_latest.sh
@@ -11,6 +11,7 @@ fi
 
 git clone git://github.com/dartsim/dart.git
 cd dart
+git checkout v6.10.1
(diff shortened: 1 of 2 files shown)


costashatz:
@matthias-mayr thanks for this. The current "documentation" is super incomplete. There is a new one coming along (see `new_docs` branch and [here](http://www.resibots.eu/robot_dart/) for a preview).

It's good to keep it a bit updated though. Thanks for this. I will merge asap..

PR #137, opened 2021-09-08
CMake Updates

costashatz:
Recently I had issues with the CMake integration and Ubuntu 20.04 (not in ArchLinux!). This PR fixes those issues.


costashatz:
I made a mistake and was using the wrong libs.. Closing this...

PR #138, opened 2021-09-15
[magnum]: do not use deprecated Magnum functions

costashatz:
Update for latest upstream Magnum...


jbmouret:
shouldn't we merge this to `devel' and keep the master to a specific version of Magnum? 

Merging this will break all the installations/containers that did not update Magnum yet, I guess...

costashatz:
We need to first have a `devel` branch.. ;)

I think that I like more the following: `master` is development, other branches or tags for stable things. I can create one for your needs. What do you think?

jbmouret:
Yes, this works for me too, but we need to keep a "release x" branch in that case :)

costashatz:
Good! I'll first make a PR to remove (the dead now) travis and use GitHub actions for CI. Then I'll create separate release branches..

costashatz:
@jbmouret I'll merge this and let me know if things will break for your dockers/setups. Of course the easiest solution is just to use latest Corrade/Magnum etc, but once we make the 1.0.0 release we'll provide installation instructions for specific commits of Corrade/Magnum etc for easy portability and avoid issues.

PR #139, opened 2021-09-22
Rotational joints error computed on the manifold

dinies:
Created class that extends PD controller to allow computation of the error of the joints on the manifold if the joints are rotational.
The approach tries to treat the angular quantities on a manifold that ensures that regions of the domain of the angular values that are locally contiguous conceptually are also contiguous numerically.

 For example when I compare two angles alfa = 355 degs and beta = -355 degs
 I want that this result  |alfa -beta| = 20 degs instead of |alfa - beta| = 710 degs.
 Same in rads alfa = 7M_PI/4, beta= 7M_PI/4,|alfa -beta| = M_PI/2  instead of |alfa -beta| = 7*M_PI/2.
This results are obtained by bounding the values of the angular quantities in a closed interval,
 in my case [-M_PI, M_PI].

### DIFF
--- a/src/robot_dart/control/manifold_control.hpp
+++ b/src/robot_dart/control/manifold_control.hpp
@@ -1,0 +1,25 @@
+#include <robot_dart/control/pd_control.hpp>
+
+namespace robot_dart {
+    namespace control {
+
+        class ManifoldControl: public PDControl {
+        public:
+
+            ManifoldControl(const std::vector<int>& t_rotational_dofs );
+            ManifoldControl(
+                const std::vector<int>& t_rotational_dofs,
+                const Eigen::VectorXd& ctrl, bool full_control = false);
+            ManifoldControl(
+                const std::vector<int>& t_rotational_dofs,
+                const Eigen::VectorXd& ctrl,
+                const std::vector<std::string>& controllable_dofs);
+
+            Eigen::VectorXd calculate(double) override;
+
+            double boxMinus(const double alfa,  const double beta);
+        private:
+            const std::vector<int> m_rotational_dofs;
+        };
+    }
+}
--- a/src/examples/franka.cpp
+++ b/src/examples/franka.cpp
@@ -1,7 +1,7 @@
 #include <iostream>
 #include <robot_dart/robot_dart_simu.hpp>
 
-#include <robot_dart/control/pd_control.hpp>
+#include <robot_dart/control/manifold_control.hpp>
 
 #ifdef GRAPHIC
 #include <robot_dart/gui/magnum/graphics.hpp>
@@ -25,8 +25,10 @@ int main()
     // set desired positions
     Eigen::VectorXd ctrl = robot_dart::make_vector({0., M_PI / 4., 0., -M_PI / 4, 0., M_PI / 2., 0., 0.});
 
+    std::vector<int> rotational_joints(ctrl.size(),1);
+
     // add the controller to the robot
-    auto controller = std::make_shared<robot_dart::control::PDControl>(ctrl);
+    auto controller = std::make_shared<robot_dart::control::ManifoldControl>(rotational_joints,ctrl);
     robot->add_controller(controller);
     controller->set_pd(300., 50.);
 
--- a/src/robot_dart/control/manifold_control.cpp
+++ b/src/robot_dart/control/manifold_control.cpp
@@ -1,0 +1,67 @@
+#include <robot_dart/control/manifold_control.hpp>
+#include "robot_dart/robot.hpp"
+#include "robot_dart/utils.hpp"
+
+namespace robot_dart {
+    namespace control {
+   
+        ManifoldControl::ManifoldControl(const std::vector<int>& t_rotational_dofs ):
+          m_rotational_dofs(t_rotational_dofs), PDControl(){}
+
+
+        ManifoldControl::ManifoldControl(
+            const std::vector<int>& t_rotational_dofs,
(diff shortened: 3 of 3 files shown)


src/robot_dart/control/manifold_control.hpp
```diff
@@ -0,0 +1,25 @@
+#include <robot_dart/control/pd_control.hpp>
+
+namespace robot_dart {
+    namespace control {
+
+        class ManifoldControl: public PDControl {
```
costashatz, line 6: The name (`Manifold`) suggests a different type of controller. I would just enhance the existing `PDControl` class.

src/robot_dart/control/manifold_control.hpp
```diff
@@ -0,0 +1,25 @@
...
+        public:
+
+            ManifoldControl(const std::vector<int>& t_rotational_dofs );
```
costashatz, line 9: Accepting a vector of `int` is inconsistent with the rest of the libraries that accepts a vector of `std::string` (for DOF names).

src/robot_dart/control/manifold_control.hpp
```diff
@@ -0,0 +1,25 @@
...
+                const std::vector<int>& t_rotational_dofs,
+                const Eigen::VectorXd& ctrl, bool full_control = false);
+            ManifoldControl(
+                const std::vector<int>& t_rotational_dofs,
+                const Eigen::VectorXd& ctrl,
+                const std::vector<std::string>& controllable_dofs);
+
+            Eigen::VectorXd calculate(double) override;
+
+            double boxMinus(const double alfa,  const double beta);
```
costashatz, line 20: Code style comment: functions are with underscores (`box_minus`). Also this should be made protected as there is no reason for it to be in the public API.

costashatz (commented):
Hello @dinies 

Thanks for this PR. I have, however, several comments (see also comments in code):

- My main comment is that we should just enhance the `PDControl` class rather than making a new one.
- My second main comment is that what you say about the angle differences is correct, BUT in all of our cases the angles are bounded (less than [-pi,pi]) so you should have never experienced an issue (except if you created a custom robot with custom limits)
- The `PDControl` class is highly incomplete, so completing it is more than welcome

Overall, what I suggest is:
- Close this PR
- Create a new branch where you edit the `PDControl` to include the corrected code
- We do not need to take as an input which joints are rotational: we can take this information from `Robot` and `Skeleton` classes (hint: we can easily check the type of joint and see if it gives angle values: `robot->dof(name)->getJoint()->getType()`)
- I like the `box_minus` function. Is there a way to do it with less computation? For example, look how I am doing this in the [Black-DROPS code](https://github.com/resibots/blackdrops/blob/master/include/blackdrops/utils/utils.hpp#L119)

Thanks again!

jbmouret:
Hey

The box minus comes from this:
```atan2(sin(x-y), cos(x-y))``` (here: https://stackoverflow.com/questions/1878907/how-can-i-find-the-difference-between-two-angles)


There is a nice alternative, I think:
```M_PI - fabs(fmod(fabs(x - y), 2*M_PI) - M_PI);```

Otherwise, I always did the for loop to get the right angle value. It is usually fast since we do just 1-2 iterations (no idea if fmod() is fast on modern CPUs...). However, I am not sure that it does what @dinies wants.

costashatz:
> There is a nice alternative, I think:
> `M_PI - fabs(fmod(fabs(x - y), 2*M_PI) - M_PI);`

I always had trouble with `fmod`. But if this works everytime, I do not have a strong opinion.

> However, I am not sure that it does what @dinies wants.

I think it does. But found creating rotation matrices a bit of an overkill for taking the difference of 2 angles. Maybe I have misunderstood. What are you trying to do @dinies?

src/robot_dart/control/manifold_control.hpp
dinies: Agreed

src/robot_dart/control/manifold_control.hpp
dinies: I will deduce internally the type of joint as you suggested

src/robot_dart/control/manifold_control.hpp
dinies: I agree

dinies:
>  What are you trying to do @dinies?
I am trying to bound the angle between [-M_PI, M_PI] in a formally descriptive way.
I am afraid that this effort of being more explicit (using the difference of rotations) may end up in suboptimal code in terms of performances. But it is something very familiar to me so I think I will try to propose that solution instead of : 
`M_PI - fabs(fmod(fabs(x - y), 2*M_PI) - M_PI)`

costashatz:
> I am afraid that this effort of being more explicit (using the difference of rotations) may end up in suboptimal code in terms of performances. But it is something very familiar to me so I think I will try to propose that solution instead of

I do not think that I prefer that explicit code if it hurts performance. This is quite easy to explain in 1 line and if it is faster, we want the fastest code. I would even be careful about `fmod` and just stick to the `while` loop if the performance is the same. The `while` loop is quite explicit, easy to understand, and if it holds the same performance as `fmod` or rotation matrices (or better), then I surely prefer the while loop.

dinies:
I have tried the other two implementations in my unit test battery and the solution with fmod does not work everytime but the other with the double while loop gives the always the expected result.
Therefore I will try to integrate that in the PD controller and make a new PR.
Thank you for your feedback.

PR #142, opened 2021-09-24
Rotational error added to pd controller

dinies:
enhanced pd_controller on error computation of angular quantities when it is dealing with revolute joints.


src/robot_dart/control/pd_control.cpp
```diff
@@ -8,6 +8,8 @@ namespace robot_dart {
         PDControl::PDControl(const Eigen::VectorXd& ctrl, bool full_control) : RobotControl(ctrl, full_control) {}
         PDControl::PDControl(const Eigen::VectorXd& ctrl, const std::vector<std::string>& controllable_dofs) : RobotControl(ctrl, controllable_dofs) {}
 
+
+
```
costashatz, line 12: Why adding spaces here?

src/robot_dart/control/pd_control.cpp
```diff
@@ -26,9 +28,21 @@ namespace robot_dart {
             Eigen::VectorXd q = robot->positions(_controllable_dofs);
             Eigen::VectorXd dq = robot->velocities(_controllable_dofs);
 
+
```
costashatz, line 31: What is the use of the space here as well?

src/robot_dart/control/pd_control.cpp
```diff
@@ -26,9 +28,21 @@ namespace robot_dart {
...
+            Eigen::VectorXd error= Eigen::VectorXd::Zero(_control_dof);
+            for(int i=0;i<_control_dof;++i){
```
costashatz, line 33: Please adhere to the formatting standard of robot_dart (use the `.clang-format` from [here](https://github.com/resibots/format_code)).

src/robot_dart/control/pd_control.cpp
```diff
@@ -26,9 +28,21 @@ namespace robot_dart {
...
+              if(robot->dof(_controllable_dofs[i])->getJoint()->getType() ==
+                  dart::dynamics::RevoluteJoint::getStaticType()){
```
costashatz, line 35: `RevoluteJoint`s are not the only ones requiring extra care:

- `BallJoint` has a full 3D orientation and uses angle axis representation
- `EulerJoint` has a full 3D orientation and uses Euler axes representation
- `FreeJoint` represents a full 3D transformation and includes an orientation

Can you do those too? I can do them if you do not feel comfortable.

By the way, we should enable/disable this with a variable (passing through the constructor, having a default value to true: i.e., using them).

src/robot_dart/control/pd_control.hpp
```diff
@@ -7,6 +7,8 @@
 #include <robot_dart/control/robot_control.hpp>
 #include <robot_dart/robot.hpp>
 
+#include <dart/dynamics/RevoluteJoint.hpp>
```
costashatz, line 10: This should go into the cpp file to not pollute the workspace of people using robot_dart.

src/robot_dart/control/pd_control.hpp
```diff
@@ -29,6 +31,16 @@ namespace robot_dart {
         protected:
             Eigen::VectorXd _Kp;
             Eigen::VectorXd _Kd;
+
+            inline double angle_dist(const double alfa, const double beta){
```
costashatz, line 35: No reason to inline this. Please put the implementation inside the cpp file. Also if you keep it as a member function, it should be prefixed by "_" (-> _angle_dist).

costashatz (commented):
Good job. Thanks for the PR. See my inline comments.

src/robot_dart/control/pd_control.cpp
dinies: From now on I will be using that formatting style

src/robot_dart/control/pd_control.cpp
dinies: This is something that I think should be handled in a different way:
since we are dealing with rotations in SE(3) with three angles I think that a method that avoids singularities of the rotation matrices is advisable.
I didn't find anything more concise than embedding the 3D rotations in quaternions and computing the error between them with the geodesic distance between them.
I am already using this in one of my modules.
If you are interested, we can try to implement this here.
Also, I am not familiar with these other types of joints and I am wondering how we can access the values of the angles:
I am used to seeing that a joint has a single value for is q.

costashatz: > This is something that I think should be handled in a different way:
> since we are dealing with rotations in SE(3) with three angles I think that a method that avoids singularities of the rotation matrices is advisable.
> I didn't find anything more concise than embedding the 3D rotations in quaternions and computing the error between them with the geodesic distance between them.

How about `dart::math::logMap(R_desired * R_current.transpose())`?

> Also, I am not familiar with these other types of joints and I am wondering how we can access the values of the angles:
> I am used to seeing that a joint has a single value for is q.

For each joint `getPositions()` gives a vector with the DOF values:

- For `BallJoint` this should be a 3x1 vector with angle-axis representation of the orientation (`dart::math::expMapRot(vector)` gives the 3x3 rotation matrix)
- For `EulerJoint` this should be a 3x1 vector with Euler angles representation (depending on the angle order [see [here](https://github.com/dartsim/dart/blob/main/dart/dynamics/EulerJoint.hpp#L95)], you can use `dart::math::eulerXXXToMatrix(vector)` [see [here](https://github.com/dartsim/dart/blob/main/dart/math/Geometry.hpp#L76)] to get the rotation matrix)
- For `FreeJoint` this should be a 6x1 vector. This can be converted to a full transformation (which then you can split into translation and orientation parts) using `dart::dynamics::FreeJoint::convertToTransform(vector)`. In the end the vector should look like [error_orientation_x, error_orientation_y, error_orientation_z, error_x, error_y, error_z].

If you are comfortable, please do this. If not, let me know and I can do this.

src/robot_dart/control/pd_control.cpp
```diff
@@ -1,12 +1,13 @@
 #include "pd_control.hpp"
 #include "robot_dart/robot.hpp"
 #include "robot_dart/utils.hpp"
+#include <dart/dynamics/RevoluteJoint.hpp>
```
costashatz, line 4: Please leave a space before the dart include..

src/robot_dart/control/pd_control.cpp
```diff
@@ -1,12 +1,13 @@
...
 
 namespace robot_dart {
     namespace control {
         PDControl::PDControl() : RobotControl() {}
-        PDControl::PDControl(const Eigen::VectorXd& ctrl, bool full_control) : RobotControl(ctrl, full_control) {}
-        PDControl::PDControl(const Eigen::VectorXd& ctrl, const std::vector<std::string>& controllable_dofs) : RobotControl(ctrl, controllable_dofs) {}
+        PDControl::PDControl(const Eigen::VectorXd& ctrl, bool full_control, const bool angular_errors) : RobotControl(ctrl, full_control), _angular_errors(angular_errors) {}
```
costashatz, line 9: No need for `const` here.

src/robot_dart/control/pd_control.cpp
costashatz: Spaces are still here..

src/robot_dart/control/pd_control.cpp
```diff
@@ -26,9 +27,21 @@ namespace robot_dart {
             Eigen::VectorXd q = robot->positions(_controllable_dofs);
             Eigen::VectorXd dq = robot->velocities(_controllable_dofs);
 
+            Eigen::VectorXd error = Eigen::VectorXd::Zero(_control_dof);
+            for (int i = 0; i < _control_dof; ++i) {
+                if (robot->dof(_controllable_dofs[i])->getJoint()->getType() == dart::dynamics::RevoluteJoint::getStaticType()
+                    and _angular_errors) {
```
costashatz, line 33: `and` is python I guess. This should never be able to compile in C++. Please merge master into this branch so that we get the CI also..

costashatz: Oh wait. No need to merge. The CI is already running. But for some unknown reason this compiled! I'll investigate this..

costashatz: Wow. This is valid in C++: https://stackoverflow.com/questions/44381428/and-keyword-in-c

Still I prefer `&&`. Can you please adapt this?

src/robot_dart/control/pd_control.cpp
dinies: I am not feeling comfortable with exponential and logarithmic maps over quaternions still.
Therefore I prefer not to dive into the implementation for the time being.

src/robot_dart/control/pd_control.cpp
dinies: done

src/robot_dart/control/pd_control.cpp
dinies: done

src/robot_dart/control/pd_control.cpp
dinies: These I am not sure I can find, can you point them out one more time ? Thank you

costashatz: @dinies do not forget this one ;)

src/robot_dart/control/pd_control.cpp
costashatz: No worries. Finish the rest and I'll do this.. ;)

src/robot_dart/control/pd_control.cpp
costashatz: > These I am not sure I can find, can you point them out one more time ? Thank you

No worries. Leave them. I'll fix them with the rotation thing..

src/robot_dart/control/pd_control.cpp
costashatz: Forgot to say thanks for the work! Thanks a lot @dinies..

dinies: Your welcome ! =)

costashatz:
@dinies can you check if the new code is working for you? I did a few tests and it seems OK for me. Once you confirm, I will merge.. Thanks!

dinies:
It works in my use case, but I only use the controller with prismatic and revolute joints so I could not say anything about the other joint controls (ballJoint, eulerJoint, freeJoint).


Speculation : Is it possible that when we are using the FreeJoint or BallJoint type (since val is of dimension six or three as opposed to a scalar value) to populate the error vector then we could incur dimensionality problems when the error is flattened (.array()) and multiplied and added to _Kd and _Kd*dq ? 
Or everything has already been thought to be consistent with the dimensions of the 'generalized dof dimension'.

costashatz:
> Or everything has already been thought to be consistent with the dimensions of the 'generalized dof dimension'.

It does not have to do with the generalized dof dimension, but with the fact that in control we can define which dimensions to control and in arbitrary order. So this means that someone can choose to re-orient the degrees of freedom of the FreeJoint or to control only one of them.

> It works in my use case, but I only use the controller with prismatic and revolute joints so I could not say anything about the other joint controls (ballJoint, eulerJoint, freeJoint).

This is enough for now.. Thanks..

costashatz:
**Merging...** Thanks a lot @dinies!

PR #152, opened 2021-09-29
Stricter warnings and avoid warnings of external libs

costashatz:
This PR cleans the code a bit by introducing stricter warnings. Surprisingly, we had very few parts in our code that needed alteration.

~~**TO-DO left:** group includes and use `#pragma GCC system_header` instead of `#pragma GCC diagnostic push/pop`.~~


costashatz:
@jbmouret @Aneoshun is this OK for you? This will make us easier to follow good coding styles for the main code of robot_dart.

PR #153, opened 2021-09-30
Allow to "break" sensors

jbmouret:
When we split a skeleton, some sensors point to unavailable bodies (or the robot would have no access to the sensors on the detached skeleton).

This PR use the `_attached` boolean (which was already here) to be able detach a sensor explicitly (to avoid a segv when it was attached to something that does not exist anymore).

One idea is that it is up to the user to know what to do when a sensor is detached (e.g., replace with zero). So, the user should check that the sensor is attached when working with damage and querying sensors.

### DIFF
--- a/src/robot_dart/sensor/sensor.cpp
+++ b/src/robot_dart/sensor/sensor.cpp
@@ -43,6 +43,14 @@ namespace robot_dart {
         void Sensor::set_pose(const Eigen::Isometry3d& tf) { _world_pose = tf; }
         const Eigen::Isometry3d& Sensor::pose() const { return _world_pose; }
 
+        void Sensor::detach() {
+            _attached_to_body = false;
+            _attached_to_joint = false;
+            _body_attached = NULL;
+            _joint_attached = NULL;
+            _active = false;
+        }
+
         void Sensor::refresh(double t)
         {
             if (_attaching_to_body && !_attached_to_body) {
--- a/src/robot_dart/sensor/sensor.cpp
+++ b/src/robot_dart/sensor/sensor.cpp
@@ -46,8 +46,8 @@ namespace robot_dart {
         void Sensor::detach() {
             _attached_to_body = false;
             _attached_to_joint = false;
-            _body_attached = NULL;
-            _joint_attached = NULL;
+            _body_attached = nullptr;
+            _joint_attached = nullptr;
             _active = false;
         }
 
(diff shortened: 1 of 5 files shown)


src/robot_dart/robot.cpp
```diff
@@ -1618,6 +1622,11 @@ namespace robot_dart {
         return detail::dof_data<15>(_skeleton, dof_names, _dof_map);
     }
 
+    Eigen::VectorXd Robot::constraint_forces(const std::vector<std::string>& dof_names) const
```
costashatz, line 1625: What does the `constraint_forces` has to do with the detaching of sensor (that this PR handles)? If this is needed for something else, can you make a different PR? Thanks!

src/robot_dart/sensor/sensor.cpp
```diff
@@ -43,8 +43,18 @@ namespace robot_dart {
         void Sensor::set_pose(const Eigen::Isometry3d& tf) { _world_pose = tf; }
         const Eigen::Isometry3d& Sensor::pose() const { return _world_pose; }
 
+        void Sensor::detach() {
+            _attached_to_body = false;
+            _attached_to_joint = false;
+            _body_attached = NULL;
```
costashatz, line 49: Can you use `nullptr` instead of `NULL`?

src/robot_dart/sensor/sensor.cpp
```diff
@@ -111,5 +120,13 @@ namespace robot_dart {
                 _attached_to_joint = false;
             }
         }
+        const std::string& Sensor::attached_to() const
+         { 
+            assert(_attached_to_body || _attached_to_joint);
```
costashatz, line 125: Please use robot_dart's assertion macros. See [here](https://github.com/resibots/robot_dart/blob/master/src/robot_dart/utils.hpp#L48-L76).

costashatz (changes requested):
@jbmouret thanks a lot for this. See my comments above.

src/robot_dart/sensor/sensor.cpp
jbmouret: sure, old habits :)

src/robot_dart/robot.cpp
jbmouret: We need it to compute the SPD+ (for position control) with a broken robot. I agree that this is not about "broken sensor", but we need broken sensors to handle damaged robots. I don't think few lines are worth a PR. We can change the commit message if you care.

src/robot_dart/sensor/sensor.cpp
costashatz: Thanks a ton!

src/robot_dart/robot.cpp
costashatz: OK then. Then change the name of the PR (e.g. `Allow to "break" sensors + constraint forces`) or put in bold in the description something about this..

costashatz (approved):

costashatz:
@jbmouret @dalinel thanks for this!

PR #154, opened 2021-10-01
fix big_sur (osx 11.6) compilation

jbmouret:
It mainly skip the check for OpenGL libs since this is a framework and Magnum already does its stuffs for this. I am not sure this works for older OS X version because it was working before without this fix. 

However, we do not have that many OS X users (except me).

### DIFF
--- a/waf_tools/magnum.py
+++ b/waf_tools/magnum.py
@@ -213,15 +213,15 @@ def check_magnum(conf, *k, **kw):
             magnum_includes = magnum_includes + [opengl_include_dir]
             conf.end_msg(opengl_include_dir)
 
-            conf.start_msg('Magnum: Checking for OpenGL lib')
             # no need to check on osx (it works anyway)
             # Osx 11.3 (Big Sur) does not have libGL.dylib anymore (there is libGL.tbd)
             # but at any rate, OpenGL is a framework so checking the dylib is not really what we need
             if conf.env['DEST_OS'] != 'darwin':
+                conf.start_msg('Magnum: Checking for OpenGL lib')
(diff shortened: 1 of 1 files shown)


waf_tools/magnum.py
```diff
@@ -214,9 +214,13 @@ def fatal(required, msg):
             conf.end_msg(opengl_include_dir)
 
             conf.start_msg('Magnum: Checking for OpenGL lib')
```
costashatz, line 216: Can you put this inside the if? No reason to say that we are looking for OpenGL if we are not.

waf_tools/magnum.py
```diff
@@ -214,9 +214,13 @@ def fatal(required, msg):
...
-            magnum_libpaths = magnum_libpaths + [opengl_lib_dir]
-            magnum_libs = magnum_libs + ['GL']
+            # no need to check on osx (it works anyway)
+            # Osx 11.3 (Big Sur) does not have libGL.dylib anymore (there is libGL.tbd)
+            # but at any rate, OpenGL is a framework so checking the dylib is not really what we need
+            if conf.env['DEST_OS'] != 'darwin':
+                opengl_lib_dir = get_directory('libGL.'+suffix, libs_check)
+                magnum_libpaths = magnum_libpaths + [opengl_lib_dir]
+                magnum_libs = magnum_libs + ['GL']
             conf.end_msg(['GL'])
```
costashatz, line 224: This also..

costashatz (changes requested):

costashatz (approved):

costashatz:
@jbmouret thanks for this.. I am merging and will check together with the Mac CI if we need to search for "frameworks".

PR #155, opened 2021-10-04
Tiago

jbmouret:
Add the Tiago.

For now, we use STL & DAE, because when I convert to .glb I do not get the right orientation (and I tried to rotate, yes).
Screen Shot 2021-10-05 at 00 23 40

TODO: we should use a box for each wheel as collision shape, so that we have fast collision.

TODO: the wheels are not working for now


costashatz:
> we use STL & DAE, because when I convert to .glb I do not get the right orientation (and I tried to rotate, yes).

I'll fix this. Do not worry.

> TODO: we should use a box for each wheel as collision shape, so that we have fast collision.

Maybe a cylinder?

> TODO: the wheels are not working for now

What do you mean are not working? There are no joints in the URDF?

jbmouret:
I don't know if the wheels are joints, but I did not try to make the base move for now (at any rate, simulating wheels with a basic cylinder is rarely perfect)

costashatz:
> For the cylinder, does the dart collision checker works with cylinders? if so, yes, it's the same.

No it does not work. But boxes will not be able to make the wheels work.

> I don't know if the wheels are joints, but I did not try to make the base move for now (at any rate, simulating wheels with a basic cylinder is rarely perfect)

They have a few examples in DART with cylinders for wheels. It will not be perfect. The other way would be to emulate the wheels with xy joints, but then we will lose interaction with the environment (e.g. going up small obstacles).

costashatz:
**Merging...**

PR #158, opened 2021-10-08
[robot]: deep copy shapes when cloning

costashatz:
This PR enables deep copying of shapes when cloning. This ensures a complete copy of the robot so that we can edit both the original and the clone completely independently (even altering shapes).

~~This is waiting https://github.com/dartsim/dart/pull/1612 before merging..~~ We enable the feature only if available (DART 6.13.0 and up)!


costashatz:
**Merging**

PR #170, opened 2022-02-10
[tiago]: Fix initial z location that ends up in collision

costashatz:


jbmouret:
I do not like to see robots falling (this is why it's good to have the robot classes ;) ).... but if I use .009 instead of 0.01, it probably collides... (slow?).

At any rate, a big source of slowness is the position of the torso_lift_joint: I think that a p=0, it collides somewhere (floor? self-collision). Things do not work with 0 (in the example, we move it with a sin(x), therefore it does not stay at 0 for long).

Also, we (I) should make a dart-collision version so that we do not have to use FCL...

costashatz:
> I do not like to see robots falling (this is why it's good to have the robot classes ;) ).... but if I use .009 instead of 0.01, it probably collides... (slow?).

If it is slow, yes it collides. I just tried and with 0.009 it is slow (this means that it collides). Also with 0.01 we do not see it falling.

> At any rate, a big source of slowness is the position of the torso_lift_joint: I think that a p=0, it collides somewhere (floor? self-collision). Things do not work with 0 (in the example, we move it with a sin(x), therefore it does not stay at 0 for long).

For me it runs quite fast.

> Also, we (I) should make a dart-collision version so that we do not have to use FCL...

Why not.. I plan on working on making Tiago movable (steering) because I need this for a few projects..

costashatz:
@jbmouret so should I merge this?

PR #173, opened 2022-02-23
cmake.in multiple inclusion

dalinel:
This PR allows the cmake.in to handle multiple inclusion case (when a library depends on several libraries that depend on  utheque / robot_dart)
Without it and in case of multiple inclusion there is an error at the  add_library stage because it has already been done

### DIFF
--- a/cmake/RobotDARTConfig.cmake.in
+++ b/cmake/RobotDARTConfig.cmake.in
@@ -4,10 +4,7 @@
 # add_executable(robot_dart_example example.cpp)
 # target_link_libraries(robot_dart_example RobotDART::Simu)
 
-if(__ROBOT_DART_CMAKE_FILE__)
-  return()
-endif()
-set(__ROBOT_DART_CMAKE_FILE__ TRUE)
+include_guard(DIRECTORY)
 
 
 include(CMakeFindDependencyMacro)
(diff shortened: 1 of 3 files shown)


dalinel:
Another way would be to use[ ```include_guard(GLOBAL)```](https://cmake.org/cmake/help/git-stage/command/include_guard.html) directly
But then we would need to ensure that the cmake version is above 3.10 with something like this:
```
if(CMAKE_VERSION VERSION_LESS 3.10)
  message(FATAL_ERROR "This file relies on consumers using CMake 3.10 or greater.")
endif()
```

jbmouret:
Don't we already require for some other deps a very recent cmake?

dalinel:
For ```RobotDart::Simu``` and ```RobotDart::Magnum```  yes 

Latest dart master has
```cmake_minimum_required(VERSION 3.10.2)```

Latest magnum master has 
```cmake_minimum_required(VERSION 3.4)```

```Utheque``` only depends on ```Boost``` so there is no cmake version requirement

jbmouret:
I think we can assume cmake > 3.10 then, no? In addition, I think 3.10 is in Ubuntu 18.04 https://packages.ubuntu.com/bionic/cmake

dalinel:
Yes I have made the changes

costashatz:
**Merging** thanks @dalinel @jbmouret

PR #174, opened 2022-03-15
[tiago]: add differential drive

costashatz:
This PR adds the possibility to Tiago to be controlled via the wheels. Here's a cool video:

https://user-images.githubusercontent.com/1283922/158372025-6322ddde-6ee3-4ee0-9f63-1f86fed33dda.mp4

Two important changes:

- URDF to allow the movement of the wheels
- Usage of bullet collision detector because FCL converts the cylinder shape to triangle mesh and this creates artifacts in the collision detection and movement

### DIFF
--- a/src/examples/tiago.cpp
+++ b/src/examples/tiago.cpp
@@ -7,9 +7,10 @@
 
 int main()
 {
-    double dt = 0.01;
+    double dt = 0.001;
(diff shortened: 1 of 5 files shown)


costashatz:
**Important:** this PR does not implement any differential drive controller (bad branch naming), but it gives the ability to do so.

PR #175, opened 2022-03-15
add -Wno-narrowing for clang options

jbmouret:
Otherwise it does not compile on the latest clang on OSX.


costashatz:
**Merging..** Ignore the error, it's from me..

PR #178, opened 2022-03-24
add a tiago with fixed wheels & casters

jbmouret:
Add the previous Tiago URDF, which is useful for whole-body control or when we want a fixed Tiago. The only difference is that the wheels and casters are 'fixed joints'.


costashatz:
@jbmouret shouldn't we add a robot class for this?

jbmouret:
I would also say that I would use the normal Tiago for simulation and this one for whole-body control. I need the URDF but not robot class.

If needed, we can use the current class with a different URDF. I would try to avoid having many different classes for variants of the same robot.

That said, if the wheels are parts of the Normal Tiago, with its own interface (which does not make sense for this specific Tiago), this could make sense to have a class.

costashatz:
OK I agree. Let's merge this...

PR #179, opened 2022-03-31
minor improvements to wscript (-fPIC) and dart.py + add line number in assert

jbmouret:
A few minor improvements:
- add `-fPIC`
- add a directory in dart.py to find dart
- add line number in ROBOT_DART_ASSERT

### DIFF
--- a/src/robot_dart/utils.hpp
+++ b/src/robot_dart/utils.hpp
@@ -51,7 +51,7 @@ namespace robot_dart {
 #define ROBOT_DART_ASSERT(condition, message, returnValue)                        \
     do {                                                                          \
         if (!(condition)) {                                                       \
-            std::cerr << "robot_dart assertion failed: " << message << std::endl; \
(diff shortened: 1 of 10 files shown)


costashatz (approved):
Two very minor comments. Other than this. Thanks! It's looking good!

costashatz:
@jbmouret this should finalized now. I did some minor modifications (e.g. when building shared, we do not need `-fPIC`; the code generated is position independent by definition) and I updated all searches for all libs (I added also search for the new M1 chips that some students of mine had problems with).. Let me know if this is working for you..

costashatz:
@jbmouret is this OK to merge?

jbmouret:
yes (in which case do we want to remove -fPIC?)

costashatz:
> (in which case do we want to remove -fPIC?)

If we are building shared libs, PIC is already there by definition so the flag is useless. Also, I cannot see an application at the moment, but maybe someone finds some weird application for removing -fPIC (e.g. an architecture/CPU not allowing PIC).

PR #182, opened 2022-05-31
[tiago]: override reset function

kounelisagis:
For Tiago, calling `reset()` sets a different orientation than the constructor. This fixes it.


jbmouret:
Thanks. This looks good to me.

costashatz:
Seems good! I'll merge and do the same for all robots..

PR #188, opened 2022-09-05
update icub.urdf

olivierrochel-inria:
* Fix joints limits to match values in xml files (motor control configuration) on the robot, and/or documentation when available.
This makes the control more consistent between the simulation and (at least) one real robot, the one at Inria Nancy. The new values have been chosen to be compatible with the ones on the robot in Nancy, and the documentation here https://icub-tech-iit.github.io/documentation/icub_kinematics/icub-joints-limits/icub-joints-limits/ 

* Change links/joints names to match the documentation and the convention used in other (official) icub models.
The names have been changed to follow the official naming convention: https://icub-tech-iit.github.io/documentation/icub_kinematics/icub-model-naming-conventions/icub-model-naming-conventions/
Only 2 joints have been renamed: `l_elbow_joint` → `l_elbow` and `r_elbow_joint` → `r_elbow` (other joints don't have the  `_joint` suffix).
Some links have been renamed to match the official doc, which sometimes results in less readable name (eg l_hip_pitch_link becomes l_hip_1 ).


jbmouret:
Thanks. I will check everything. I think we should rename the robot as icub_nancy01.urdf as this is very specific to our robot. What do you think?

costashatz:
> I think we should rename the robot as icub_nancy01.urdf as this is very specific to our robot. What do you think?

I agree although the current iCub.urdf was built around the Nancy robot. So we could rename the robot as a whole? Or just add a note in the docs/files?

jbmouret:
besides this naming question, I think we should merge asap.

olivierrochel-inria:
I'd :+1: the "note in the docs/files" option: people who own an icub know there's no "standard" icub so they'll look for that piece of information anyway, for the others that's not critical. I'm not against renaming the robot though if you think it's clearer that way.

costashatz:
I think we should merge.. No? @olivierrochel-inria @jbmouret ?

olivierrochel-inria:
I think we can merge, yes. As for the name the doc says "There are currently 42 iCUbs in the world, and many versions. [...] Our model is close to the Inria's iCub" so the information is there, we can see later if it needs to be made more explicit.

jbmouret:
Yes good for me.

PR #190, opened 2022-09-26
Removing Boost as a dependency

costashatz:
As of DART v6.12.2, Boost is no longer a dependency for DART. Thus, there is no reason for robot_dart to have Boost as a dependency. This PR replaces every usage of Boost with C++17 or custom made functions/libs.


costashatz:
@jbmouret please check this and let me know if this runs in your setup.. With DART v6.12.2 we should be completely Boost-free!

jbmouret:
I think this does not work on our Ubuntu 18.04 (current Ubuntu on PAL robots...) because clang-10 apparently does not implement all the c++17 standard library. Maybe we could have a few #ifdef for keeping the old code (but I don't know if the compiler has a define for this).

costashatz:
Interesting.. According to clang (see [here](https://clang.llvm.org/cxx_status.html)), clang supports C++17 from clang5. What is the output of `llvm-config --version`?

costashatz:
So it turns out that Ubuntu 18.04 come with the libstc++ version of gcc7. gcc7 is a weird compiler version for C++17. It supports all features of C++17, but the `std::filesystem` is at `<experimental/filesystem>` and we need to link to a library (`stdc++fs`). The last commit should handle those cases and propagate everything also to the CMake files. @jbmouret let me know if this works for you.

costashatz:
@jbmouret any updates on this one? Thanks!

PR #192, opened 2023-03-30
waf_tools/magnum: fix libdl detection

olivierrochel-inria:
fix libdl detection for magnum in waf_tools when libdl.so is found but not libdl.so.2, that was breaking the build in ubuntu18.04 at least.

### DIFF
--- a/waf_tools/magnum.py
+++ b/waf_tools/magnum.py
@@ -354,6 +354,7 @@ def check_magnum(conf, *k, **kw):
                                     magnum_component_libs[component].append('dl')
 
                                     glfw_found = True
+                                    break
                                 except:
                                     glfw_found = False
 


costashatz:
Thanks! I'll merge as soon as the CI tests pass.

olivierrochel-inria:
ok if I understand correctly configure fails in CI when trying to find dartpy, with gcc only. Is this a known issue or should I investigate ? On my side building is fine with ubuntu 20.04 and 18.04 (docker out of the box, come custom install paths, gcc, python ON)

costashatz:
> Is this a known issue or should I investigate ?

It's not a known issue, but my other branches failed there as well (tried yesterday). So it's not related to your PR. **Merging** and I will investigate later.

Issue #201, opened 2024-06-17
"One line installation" fails on a vanilla MACOS 14 install


c-joly:
Hello @costashatz, this is an issue linked to the JOSS review https://github.com/openjournals/joss-reviews/issues/6771

I tried to install robot_dart on a fresh install of MacOS 14.5 (Virtual Machine, M2 series Processor, MACOS 14.5), but it did not succeed. I attached the terminal output at https://gist.github.com/c-joly/e26956198d2438ef62f1291c0b6c2809. Here are the main problems that I see (but it might have others):
  - `pip3 install dartpy` does not pass
  - `cmake` is not found, so no Makefile is generated and the make command provided no output
  -  `waf` not found
  - After installing `cmake` via `brew`, there are other problems with pip3 that appears obviously ("This environment is externally managed; To install Python packages system-wide, try brew install...").
  - I also saw an error `python not found`... By default, the system uses `python3` command only, there is no simlink `python` on a my vanilla MACOS-14.

It looks like the one line installation script makes several hypotheses concerning the machine configuration (`homebrew`, `cmake`, some installation of python). **Can you list the requirements?** 

Finally, I realized that in case of error, the script continues to execute, so it is difficult to know if the installation succeeded or not (it might be possible to find some errors in the middle of the log). I suggest to configure the script **to stop at the first encountered error** (see `bash strict mode`). When it is configured like this, it stops with the error `python not found`.

costashatz:
Hello @c-joly,

Thanks for the comment and the effort.

> I attached the terminal output at https://gist.github.com/c-joly/e26956198d2438ef62f1291c0b6c2809

Thanks for this. Really helpful.

> `pip3 install dartpy` does not pass

Hmmm. I used pip to avoid having to deal with compiling from source. I am suspecting that the dart developers do not have pre-built binaries for M2 processors. I will update the script to build DART from source.

> `cmake` is not found, so no Makefile is generated and the make command provided no output

This is very weird. `brew install dartsim --only-dependencies` and `brew install binutils pybind11` should install `cmake`! In my MacOS CI (MacOS 12), I am not installing `cmake` explicitly and it works. In any case, this is an easy fix: I will update the script to install all the base compilation packages (clang, cmake, etc).

> `waf` not found

Where did you run the script from? You should be in the root of the robot_dart repo. I guess this might be related to the python issue. *Ahh no*, I see why this happened. I will fix it.

[...]

PR #206, opened 2024-06-28
Added install script for Mac Silicon which compiles dart

c-joly:
- This script is based on the install_osx script. It compiles and install `dart` / `dartpy` following the steps provided on the documentation.

- An ugly patch has been added to include <std.io> in a header (maybe it can be avoided if we force cmake to use g++ instead of clang which is used by default?)

- Some environment variables have been changed/added (PYTHONPATH, DYLD_LIBRARY_PATH)

- **I did not add any additional installation in the script** (like `cmake`, `xquartz` or `ffmpeg`). These commands are all installable with `homebrew` but they might have been already installed by the user by other means (e.g. `cmake` and `xquartz` have they own PKG/DMG installers). To avoid any side effect to the user, I did not add any of these install in the script, but documentation should mention these programs as prerequisite.

### DIFF
--- a/scripts/install_silicon.sh
+++ b/scripts/install_silicon.sh
@@ -1,0 +1,134 @@
+set -e
+
+CLEAN=0
+
+CLEAN=${1:-$CLEAN}
+
+brew install binutils pybind11
+python -m pip install --upgrade pip
+pip3 install numpy
+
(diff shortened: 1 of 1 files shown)


costashatz:
This is fixed! See #201!

